PR #43, opened 2021-09-23
Work in Progress.

Ksteeds:
Need to complete but adding the foundation level code. Finishing off in the next couple of hours.


Ksteeds:
If XCATS are xcats = ['XR', 'CRY', 'GROWTH', 'INFL'], the weighting functionality will fail on 'XR' as its recorded over the longest time period. Just need to make a few minor changes and subsequently should be complete.

rsueppel (approved):

PR #68, opened 2021-10-06
Add interface to DataQuery

lsimonsen:


lsimonsen:
@Ksteeds please review so we can merge in, once merged in we can download data from DQ such as:

```python
import yaml
from macrosynergy.management.dq import DataQueryInterface


if __name__ == "__main__":
    with open("../config.yml", 'r') as f:
        cf = yaml.load(f, Loader=yaml.FullLoader)
    print(cf)
    with DataQueryInterface(
            username=cf["dq"]["username"],
            password=cf["dq"]["password"],
            crt="../api_macrosynergy_com.crt",
            key="../api_macrosynergy_com.key"
    ) as dq:
        print("Connected:", dq.check_connection())

        r = dq.get_ts_expression(
            expression="DB(JPMAQS,BRL_CPIXFE_SJA_P3M3ML3AR,value)",
            start_date="1999-01-01"
        )
        print(r)
```

PR #102, opened 2021-11-10
Feature/dq_simplification

Ksteeds:
Minor updates but also important updates. 

Removed the print() statement which I believe was causing the warning when run in a Jupyter Notebook.


Ksteeds:
Submitted a Pull Request to address a relatively subtle bug - existing logic is all correct but required a minor modification. Still working on the Jupyter file.

PR #157, opened 2021-12-06
Redesign of code.

Ksteeds:
Requires further discussion.

### DIFF
--- a/macrosynergy/management/dq.py
+++ b/macrosynergy/management/dq.py
@@ -325,35 +325,32 @@ class DataQueryInterface(object):
         params.update(params_)
 
         no_tickers = len(tickers)
+        print(f"Number of tickers: {no_tickers}.")
         iterations = ceil(no_tickers / 20)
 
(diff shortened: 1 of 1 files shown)


Ksteeds:
Side note the algorithm still requires changing the time delay away from a fixed parameter.

macrosynergy/management/dq.py
```diff
@@ -325,35 +325,32 @@ def _request(self, endpoint: str, tickers: List[str], params: dict,
...
+                        raise ValueError("Server being hit too quickly with requests.")
+                    else:
+                        if isinstance(response, list):
+                            final_output.extend(response)
                         else:
-                            if isinstance(response, list):
-                                final_output.extend(response)
-                            else:
-                                continue
+                            continue
```
lsimonsen, line 353: I think this else continue statement is redundant, as it is at the end of the for loop?

lsimonsen (approved):

PR #164, opened 2021-12-09
Reversion to previous framework. Works consistently.

Ksteeds:
Code works & is robust. Will require further performance reviews.

But the average response time for the below request is around the 4 minute mark.

[2020219 rows x 4 columns]
Unique tickers in the df: 353.


Ksteeds:
I'll look to debug the current framework over the coming days but in terms of performance the response time is almost identical, so the difference is more stylistically.

lsimonsen (approved):

PR #188, opened 2022-01-04
Accounts for inconsistent data across metrics.

Ksteeds:
Modifications made if a single metric is missing but the residual metrics are available.


Ksteeds:
However, if the 'value' metric is missing, on a particuliar ticker, the ticker will not be included in the returned dataframe.

rsueppel (approved):

PR #199, opened 2022-01-10
Temporary submission prior to more rigorous error catching mec…

Ksteeds:
…hanism.

Work-in-progress. Awaiting response from DQ.


Ksteeds:
Remain in testing mode.

PR #244, opened 2022-02-01
Bugfix/target positions

Ksteeds:
Removal of xcat parameter from the source code.


Ksteeds:
The only aspect of uncertainty is the application of blacklist period.

rsueppel (approved):

PR #321, opened 2022-04-22
Feature/view_timelines

Ksteeds:
Handling for Seaborn's default linear interpolation.


Ksteeds:
Work incomplete.

Ksteeds:
Issue addressed. However, graphs take longer to be produced.

Additionally, potential issue with cumulative sum graph: break results in different colour line.

Ksteeds:
Requires a computationally faster solution.

PR #341, opened 2022-05-06
Feature/signal_return

Ksteeds:
Erroneous values handled.


Ksteeds:
Code has now handled for the falsified zeros (misleading zeros which should have been NaNs and subsequently removed from the DataFrame).

rsueppel (approved):

Issue #342, opened 2022-05-09
make_zn_score: sequential estimation needs to increase speed with added parameter


rsueppel:
At the moment choosing sequential=True makes the function too slow for large panels.
The solution could be an added parameter est_freq="m", with options "w", "q" and "a" that govern the frequency at which the means and standard deviations are estimated (and subsequently forward-filled.

Beyond, the function also runs very slow when `sequential=False` is chosen, which suggests that the calculation is very inefficient.

Ksteeds:
Tied to the latest Pull Request.

PR #352, opened 2022-05-16
feature/make_zn_scores_efficient

Ksteeds:
def func_executor(df: pd.DataFrame, neutral: str, n: int):
    """
    Function used to clean up the repetitive code in the below methods. Will produce the
    evolving neutral level.

    :param <pd.DataFrame> df:
    :param <str> neutral:
    :param <int> n: number of dates the neutral level is computed over.
    """

    if neutral == "mean":
        return np.array([df.iloc[0:(i + 1), :].stack().mean() for i in range(n)])
    else:
        return np.array([df.iloc[0:(i + 1), :].stack().median() for i in range(n)])


Ksteeds:
Tied to the issue #342.

Ksteeds:
Time taken for functional executor, mean: 9.492730140686035
Overall time taken for make_zn_scores(): 19.86984610557556

Definitely scope to improve the above computational speed of a rolling mean.

Ksteeds:
Following upgrades median algorithm takes: Time taken: **11.616146087646484**.
Following upgrades mean algorithm takes: Time taken: **12.103456974029541**

Ksteeds:
def iis_std_panel(dfx: pd.DataFrame, min_obs: int, sequential: bool = True,
                  iis: bool = True):
    """
    Function designed to compute the standard deviations but accounts for in-sampling
    period. The in-sampling standard deviation will be a fixed value.
    """
Aim to improve the performance speed of this subroutine.

Ksteeds:
Code refactoring complete for down-sampling.

PR #358, opened 2022-05-21
make_zn_score: Docs changes and Todos

rsueppel:


macrosynergy/panel/make_zn_scores.py
```diff
@@ -389,6 +410,8 @@ def make_zn_scores(df: pd.DataFrame, xcat: str, cids: List[str] = None,
 
     if est_freq != "d":
         dfw = dfw.resample(pd_freq[est_freq], axis=0, convention='end').last()
+        # Todo: Fatal bug. This means subsequent applications of pen_neutral etc.
```
Ksteeds, line 413: Is this not the purpose ? To calculate the neutral levels and standard deviations on a lower frequency. For instance, an assert that has a low variance which will subsequently not fluctuate a considerable amount over a week.

macrosynergy/panel/make_zn_scores.py
```diff
@@ -26,50 +28,57 @@ def pan_neutral(df: pd.DataFrame, neutral: str = 'zero', sequential: bool = Fals
...
+            # Todo: incorrect as min_obs is defined on days, while dfw could be weeks
+            #  or months.
+            # Todo: calculate earliest date based on days (timedelta) and use
 
         elif sequential and iis:
             iis_neutral = np.repeat(df.iloc[0:min_obs].stack().mean(),
                                     min_obs)
 
             os_neutral = rolling_mean_with_nan(dfw=df)
+            # Todo: incorrect, use pandas df.expanding.mean()
```
Ksteeds, line 76: Why incorrect ? I'm not sure df.expanding.mean() works when computing a rolling mean across a panel.

PR #361, opened 2022-05-27
feature/dq_oauth

Ksteeds:
OAuth.

Still requires testing and integration into DQ.py file.

### DIFF
--- a/macrosynergy/management/OAuth_connection.py
+++ b/macrosynergy/management/OAuth_connection.py
@@ -1,0 +1,124 @@
+
+"""
+DataQuery Interfaces and API wrappers.
+"""
+from oauthlib.oauth2 import BackendApplicationClient
+from requests_oauthlib import OAuth2Session
+import json
+import logging
+from typing import Union
+# The functools module is for higher-order functions: functions that act on or return
+# other functions.
+from functools import partial
+
+logger = logging.getLogger(__name__)
+
+BASE_URL = "https://api-developer.jpmorgan.com/research/dataquery-authe/api/v2/"
+
+class DataQueryWebAPI(object):
+    """
+    DataQuery REST web API class
+
+    For unittesting mock up methods
+
+        1. _create_api (return: mp_api)
+        2. _renew_token (mb_api fetch_token)
+        3. _fetch (mb_api get and post)
+
+    :param <str> client_id: string with client id, username.
+    :param <str> client_secret: string with client secret, password.
+    """
+
+    def __init__(self, client_id: str, client_secret: str, token: str):
+        logger.info("Establish connection to DataQuery Web API")
+        self.__is_connected: bool = False
+
+        self.__token_url = "https://authe.jpmchase.com/as/token.oauth2"
+        self.__dq_api_resource_id = 'JPMC:URI:RS-06785-DataQueryExternalApi-PROD'
+
+        id_error = f"client_id argument must be a string and not <{type(client_id)}>."
+        assert isinstance(client_id, str), id_error
+        self.client_id: str = client_id
+
+        secret_error = f"client_secret must be a str and not <{type(client_secret)}>."
+        assert isinstance(client_secret, str), secret_error
+
+        self.client_secret: str = client_secret
+
+        self.mb_api = self._create_api()
+        if not self.mb_api.authorized:
+            self._renew_token()
+
+    def _set_is_connected(self, connected: bool):
+        self.__is_connected = connected
+
+    def is_connected(self):
+        return self.__is_connected
+
+    def _create_api(self):
+        client = BackendApplicationClient(client_id=self.client_id)
+        mb_api = OAuth2Session(client=client)
+        return mb_api
+
+    def _renew_token(self):
+        self.mb_api.fetch_token(
+            token_url=self.__token_url,
+            client_id=self.client_id,
+            client_secret=self.client_secret,
+            aud=self.__dq_api_resource_id
+        )
+        self._set_is_connected(connected=self.mb_api.authorized)
+
+    def _fetch(self, request: str, method: str = "POST",
+               payload: Union[list, dict] = None, params: dict = None):
+        """
+
+        :param <str> request: string with request elements and parameters.
+        :param <str> method: default 'GET', either 'GET' or 'POST' http methods.
+        :param <list/dict> payload: default None, defines payload for a 'POST' method.
+
+        :return: list with response from DataQuery Web API.
+        :rtype: <list[dict]>
+        """
+
+        assert isinstance(request, str), f"request of unknown type {type(request)}: {request}"
+
+        assert isinstance(
+            method, str
+        ), f"method of unknown type {type(method)}: {method}"
+
+        # Helper method that will request new access token if required
+        url = BASE_URL + request
+
+        # This is typically the first time when we do not yet have a token
+        if not self.mb_api.authorized:
+            self._renew_token()
+
+        if method == "GET":
+            call_url = self.mb_api.get
+
+        elif method == "POST":
+            headers = {"Content-Type": "application/json"}
+            call_url = partial(self.mb_api.post, headers=headers, params=params,
+                               data=json.dumps(payload))
+        else:
+            raise NotImplementedError(f"Unknown method; {method:s}")
+
+        try:
+            r = call_url(url)
+        except:
+            self._renew_token()
+            r = call_url(url)
+
+        if r.status_code == 401:
+            # If authorization failed, it is likely that the token has expired.
+            # Get a new one and try again.
+            self._renew_token()
+            r = call_url(url)
+
+        error_status = f"Status code {r.status_code} for request {request:s}"
+        assert (r.status_code == 200), error_status
+
+        json_data = json.loads(r.content)
+
+        return json_data


Ksteeds:
Requires testing.

macrosynergy/management/OAuth_connection.py
```diff
@@ -0,0 +1,98 @@
...
+    :param <str> client_secret: string with client secret, password.
+    """
+
+    def __init__(self, client_id: str, client_secret: str):
+
+        url = "https://api-developer.jpmorgan.com/research/dataquery-authe/api/v2/"
+        self.base_url = url
+
+        self.__token_url = "https://authe.jpmchase.com/as/token.oauth2"
+        self.__dq_api_resource_id = 'JPMC:URI:RS-06785-DataQueryExternalApi-PROD'
```
lsimonsen, line 33: Similar to above `url` for `TOKEN_URL = "https://authe.jpmchase.com/as/token.oauth2"` and `DQ_RESOURCE_ID = "JPMC:URI:RS-06785-DataQueryExternalApi-PROD"`

lsimonsen:
As discussed @Ksteeds here is the propose re-working of the folder structure for the DataQuery interface:

* macrosynergy
  * dataquery
    * api.py
      * `Interface` class
    * auth.py 
      * `OAuth` class for OAuth authentication
      * `CertAuth` class for certificate authentication

To keep back-ward compatibility we can rework the current DataQuery interface file [dq.py](https://github.com/macrosynergy/macrosynergy/blob/develop/macrosynergy/management/dq.py) into:
```python
import warnings
from ..dataquery.api import Interface as DataQueryInterface

warnings.warn(
    "DataQueryInterface has been moved to macrosynergy.dataquery.api"
   DeprecationWArning,
   stacklevel=2
)

__all__ = ["DataQueryInterface"]

```
This should insure old code will still work.

macrosynergy/management/certificate_connection.py
```diff
@@ -0,0 +1,113 @@
...
+    :param <str> key: string with private key location.
+    """
+
+    def __init__(self, username: str,
+                 password: str,
+                 crt: str = "api_macrosynergy_com.crt",
+                 key: str = "api_macrosynergy_com.key"):
+
+        u_error = f"username must be a <str> and not <{type(username)}>."
+        assert isinstance(username, str), u_error
```
lsimonsen, line 30: Similar to above:
```python
assert isinstance(username, str), (
    f"username must be a <str> and not <{type(username)}>."
)
```

macrosynergy/management/certificate_connection.py
```diff
@@ -0,0 +1,113 @@
...
+
+        u_password = f"password must be a <str> and not <{type(password)}>."
+        assert isinstance(password, str), u_password
```
lsimonsen, line 33: And again:
```python
assert isinstance(password, str), (
    f"password must be a <str> and not <{type(password)}>."
)
```

macrosynergy/management/certificate_connection.py
```diff
@@ -0,0 +1,113 @@
...
+        """
+        Validates the key & certificate exist in the referenced directory.
+        """
+        error_type = file_type + " file must be a <str>."
+        assert isinstance(directory, str), error_type
+
+        try:
+            os.path.exists(directory)
+        except OSError as err:
+            print("OS error: {0}".format(err))
```
lsimonsen, line 57: Why are we ignoring `OSError`? Should we not still raise it again or handle it?

lsimonsen: Currently, if I am not mistaken we return a `None` value from `valid_path` in case of `OSError` - is this desirable behaviour?

macrosynergy/management/certificate_connection.py
```diff
@@ -0,0 +1,113 @@
...
+            return directory
+
+    def get_dq_api_result(self, url, params: dict = None):
+        """
+        Method used exclusively to request data from the API.
+        """
+        r = requests.get(url=url, cert=(self.crt, self.key),
+                         headers=self.headers, params=params)
+
+        return r
```
lsimonsen, line 68: Similar to above for OAuth, use context `with` statement, and also consolidate with `_fetch`, doesn't seem like we need both functions (unless we want fetch back in the `DataQueryInterface` itself and it is shared code between OAuth and Certificate Authentication).

macrosynergy/management/certificate_connection.py
```diff
@@ -0,0 +1,113 @@
...
+
+        url = self.base_url + endpoint
+        self.last_url = url
+
+        results = []
+
+        auth_check = lambda string: string.split('-')[1].strip().split('<')[0]
+
+        r = self.get_dq_api_result(url=url, params=params)
+        self.status_code = r.status_code
```
lsimonsen, line 92: Check status code is okay:
```python
assert r.ok
```

macrosynergy/management/certificate_connection.py
```diff
@@ -0,0 +1,113 @@
...
+        self.last_response = r.text
+
+        if self.last_response[0] != "{":
```
lsimonsen, line 95: We should be able to get this form the `status_code` 401: "The user has not successfully authenticated with the service." from below.

macrosynergy/management/OAuth_connection.py
```diff
@@ -0,0 +1,98 @@
+
+"""
+DataQuery Interfaces and API wrappers.
+"""
+import requests
+import json
+from typing import Union
+# The functools module is for higher-order functions: functions that act on or return
+# other functions.
+from functools import partial
```
lsimonsen, line 10: Are we using this? Remove?

macrosynergy/management/dq.py
```diff
@@ -5,28 +5,19 @@
...
-BASE_URL = "https://platform.jpmorgan.com/research/dataquery/api/v2"
-
-DQ_ERROR_MSG = {
-    204: "Content requested unavailable.",
-    400: "Request but it was malformed or invalid.",
-    401: "The user has not successfully authenticated with the service.",
-    500: "There was an internal server error."
-}
+from macrosynergy.management.OAuth_connection import DataQueryOAuth
+from macrosynergy.management.certificate_connection import DataQueryCert
```
lsimonsen, line 20: With new package structure, we can refactor this into:
```python
from .auth import OAuth, CertAuth
```

macrosynergy/management/dq.py
```diff
@@ -236,8 +155,10 @@ def _fetch_threading(self, endpoint, params: dict):
 
         while clause(counter):
```
lsimonsen, line 156: Why are we defining a clause lambda function that hard-codes a max of 5 re-fetching?

Either it should be:
```python
while True:
   # TODO break point explicitly built into code
   pass
```
OR we should have it as a configurable state variable, if we have reason to believe there is a fixed set of responses:
```python
while counter <= 5:
    pass
```

lsimonsen: Got it, the counter is for the connection error/timeout. In that case I would skip the lambda function and write it directly as: `while counter <= 5` to make it clearer.

lsimonsen (changes requested):

macrosynergy/management/OAuth_connection.py
```diff
@@ -0,0 +1,98 @@
...
+                         'expires_in': json['expires_in']}
+
+        return json['access_token']
+
+    def get_dq_api_result(self, url, params: dict = None):
+        """
+        Method used exclusively to request data from the API.
+        """
+        r = requests.get(url=url, params=params,
+                         headers={'Authorization': 'Bearer ' + self._get_token()},
```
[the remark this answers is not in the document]
Ksteeds: Subtle aspects of the code.

The method _fetch(), in both the certificate and OAuth Class, is only used to check whether the connection to DQ is stable. Therefore, the while loop, to request all the data, has been excluded. That method is only used for the purpose of testing connection: check_connection() method.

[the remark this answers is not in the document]
Ksteeds: Additionally, the while loop has been included in the method _fetch_multithreading() which is the only "fetch" method involved in returning the data. Therefore, I believe the implementation is correct.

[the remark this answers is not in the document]
Ksteeds: get_dq_api_result() method is required to support _fetch_multithreading(). _fetch_multithreading() requires the request() but the approach varies across certificate and OAuth. Hence, it is held in a separate method (required because of the difference).

macrosynergy/management/dq.py
```diff
@@ -71,46 +53,31 @@ class DataQueryInterface(object):
...
-        self.crt = crt
+                 thread_handler: int = 20,
+                 **kwargs):
+
+        if oauth:
+            client_id = kwargs.pop('client_id')
+            client_secret = kwargs.pop('client_secret')
+            access = DataQueryOAuth(client_id=client_id,
+                                    client_secret=client_secret)
+            self.access = access
```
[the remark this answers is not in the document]
Ksteeds: Stylistic preference. Either or.

macrosynergy/management/dq.py
```diff
@@ -71,46 +53,31 @@ class DataQueryInterface(object):
...
+        else:
+            username = kwargs.pop('username')
+            password = kwargs.pop('password')
+            crt = kwargs.pop('crt')
+            key = kwargs.pop('key')
+            access = DataQueryCert(username=username, password=password,
+                                   crt=crt, key=key)
+            self.access = access
```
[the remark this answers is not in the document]
Ksteeds: Similar to above.

macrosynergy/management/dq.py
Ksteeds: Changed.

Included to try the four possible servers. If there is not a response after five tries, able to deduce that the servers are down.

macrosynergy/management/OAuth_connection.py
```diff
@@ -0,0 +1,98 @@
...
+
+        results = []
+        url = self.base_url + endpoint
+        r = self.get_dq_api_result(url=url, params=params)
+
+        self.last_response = r.text
+        response = json.loads(self.last_response)
+
+        assert select in response.keys()
+        results.extend(response[select])
```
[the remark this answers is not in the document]
Ksteeds: See below.

In the method _fetch_threading() this is used.

_fetch() method is used to check the connection to DQ. Hence, not required.

macrosynergy/management/OAuth_connection.py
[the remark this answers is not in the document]
Ksteeds: Regarding re-using the token - it was adding overhead but now the token will be used continuously until it has expired. Response time now more closely aligned with the certificate approach.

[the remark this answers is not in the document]
Ksteeds: Resolving given changes have been made & committed.

macrosynergy/management/OAuth_connection.py
```diff
@@ -0,0 +1,98 @@
...
+
+        json = requests.post(url=self.__token_url,
+                proxies= {},
+                data={'grant_type': 'client_credentials', 'client_id': self.client_id,
+                      'client_secret': self.client_secret,
+                      'aud': self.__dq_api_resource_id}).json()
+
+        _stored_token = {'created_at': datetime.now(),
+                         'access_token': json['access_token'],
+                         'expires_in': json['expires_in']}
```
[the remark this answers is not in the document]
Ksteeds: Originally the solution was naively requesting a new token each time which slowed the response time. Changes have been made.

PR #363, opened 2022-05-28
naive_pnl: TODOs

rsueppel:
Benchmark series must be taken from a JPMaQS dataframe, not a separate series.
Also, they can sit comfortably in the main dataframe (self.df).
Indeed, we may call add_bm right at initialization. We just need to be careful that we do not duplicated return series, in case the benchmark is also a target return.


Ksteeds:
Integrated changes into parent branch.

PR #391, opened 2022-06-20
ToDos for make_zn_score

rsueppel:


Ksteeds:
First error potentially applicable if blacklist period is applied from the series' start date - leading NaN timestamps might inadvertently be removed.

Ksteeds:
Solved in separate PR. Closing.

PR #396, opened 2022-06-21
feature/signal_returns_unit

Ksteeds:
Unit Test.


Ksteeds:
Parameter "type" should be modified to "cs_type" in future.

rsueppel (approved):

PR #410, opened 2022-07-05
bugfix/panel_calculator

Ksteeds:
Subtle bug.


Ksteeds:
Believe the origination of the intersection bug is caused by:

    dfx = reduce_df(df, xcats=old_xcats_used, cids=cids,
                    start=start, end=end, blacklist=blacklist,
                    intersect=True)

Ksteeds:
Believe second bug regarding a category ending with a digit has also been resolved.

rsueppel (approved):

PR #434, opened 2022-07-14
feature/multiple_signals

Ksteeds:
Todo: add method .rival_sigs(sigs, sig_neg: bool = False)
In this case a comparative output table is returned with the Panel
row of the original table only, but for the original signal and
all rival categories.

### DIFF
--- a/macrosynergy/signal/signal_return.py
+++ b/macrosynergy/signal/signal_return.py
@@ -57,19 +57,30 @@ class SignalReturnRelations:
                         'pos_prec', 'neg_prec', 'pearson', 'pearson_pval',
                         'kendall', 'kendall_pval']
 
-        if sigs is not None:
+        self.ret = ret
+        self.freq = freq
 
-            assert isinstance(sigs, list), "List of signal(s) expected."
-            xcats = list(df['xcat'].unique())
-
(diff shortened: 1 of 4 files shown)


Ksteeds:
First proposal is complete.

macrosynergy/signal/signal_return.py
```diff
@@ -33,46 +34,117 @@ class SignalReturnRelations:
...
+        if rival_sigs is not None:
+
+            if isinstance(rival_sigs, str):
+                rival_sigs = [rival_sigs]
+            assert isinstance(rival_sigs, list), "List of signal(s) expected."
+            xcats = list(df['xcat'].unique())
+
+            df_sigs = self.__signal_table__(df=dfd, sigs=rival_sigs, xcats=xcats,
+                                            agg_sig=agg_sig)
+            # Todo: dfd has nor reference, should this be df?
```
Ksteeds, line 85: Typo - should be "df". Problem with Python that it will still reference the variable from the parent scope, as opposed to the locally passed DataFrame.

Ksteeds:
Update of Unit Test still required.

rsueppel (approved):

PR #477, opened 2022-09-06
updated server error message

GlennRe9:
Updated server error message explaining it has to do with the absence of instruments in the response json

### DIFF
--- a/macrosynergy/dataquery/api.py
+++ b/macrosynergy/dataquery/api.py
@@ -109,7 +109,7 @@ class Interface(object):
         try:
             response[select]
         except KeyError:
-            print(f"{response['errors'][0]['message']} - will try a different server.")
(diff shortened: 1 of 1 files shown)


macrosynergy/dataquery/api.py
```diff
@@ -109,7 +109,7 @@ def server_retry(response: dict, select: str):
         try:
             response[select]
         except KeyError:
-            print(f"{response['errors'][0]['message']} - will try a different server.")
+            print(f"{response['errors'][0]['message']} no instruments in response - will try a different server.")
```
lsimonsen, line 112: I would suggest we write it as:
```python
print(f"Key {select} not found in response: {response} --- retry download.")
```

Ksteeds: Transferring to an issue along with setting the delay parameter's default value to zero. Will close PR. Housekeeping.

PR #500, opened 2022-09-21
bugfix/view_ranges

Ksteeds:
Minor bug in view_ranges.

Bug with sort_cids_by if set to None. Conditional statements required a change.


rsueppel (dismissed):

Ksteeds:
Problem still persists. Will revert to the original solution.

lsimonsen (approved):

PR #502, opened 2022-09-21
feature/setup_18

Ksteeds:
PyPi 0.0.18


Ksteeds:
Will fail until https://github.com/macrosynergy/macrosynergy/pull/500 has been approved.

Ksteeds:
Problem still persists. Will revert to the original solution.

lsimonsen (approved):

PR #527, opened 2022-09-29
Adding feature to publish to TestPyPI

Magnus167:
Adding a feature to publish to TestPyPI using GitHub Actions


tests/integration/test_testpypi_version.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+class TestPyPIVersionNumber(unittest.TestCase):
+
+    def test_version_number(self):
+
+        # test for TestPyPI version number
+
+        package = 'macrosynergy'
+        with requests.get(f'https://test.pypi.org/pypi/{package}/json') as r:
```
lsimonsen, line 23: Why not put the `package` name directly in the URL, so that it becomes `with requests.get(f'https://test.pypi.org/pypi/macrosynergy/json') as r:`?

.github/workflows/python-test-pypi-package.yml
```diff
@@ -0,0 +1,75 @@
+# This is a basic workflow to help you get started with Actions
+
+name: Test & Upload to TestPyPI
+
+# Controls when the action will run. 
+on:
+  # Triggers the workflow on push to the test branch
+  push:
```
lsimonsen, line 8: Just to double check, does PUSH include raising a PR?

setup.py
```diff
@@ -34,8 +34,10 @@
 
 MAJOR = 0
 MINOR = 0
+
 MICRO = 18
 ISRELEASED = True
+
```
lsimonsen, line 40: Any particular reasons for the blank spaces?

lsimonsen (changes requested):
@Magnus167 and @mikiinterfiore left a few comments for you. Let me know what you think.

.github/workflows/python-test-pypi-package.yml
Magnus167: I missed that; there is a separate event for `on: pull_request: [closed]`, which is followed by an `if_merged` job in the examples. I'm looking into this, and trying to figure out how to run the job if the `test` branch is target of the pull request.

Source : [GitHub Actions/Using workflows/Events that trigger workflows/Running your workflow when a pull request merges](https://docs.github.com/en/github-ae@latest/actions/using-workflows/events-that-trigger-workflows#running-your-workflow-when-a-pull-request-merges)

tests/integration/test_testpypi_version.py
```diff
@@ -0,0 +1,34 @@
...
+# already part of python
+
+try:
+    from .version import git_revision as __git_revision__
+    from .version import version as __version__
+except ModuleNotFoundError:
+    from setup import get_version_info
+    FULLVERSION, GIT_REVISION = get_version_info()
+    __version__ = FULLVERSION
+    __git_revision__ = GIT_REVISION
```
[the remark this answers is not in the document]
Magnus167: Updated that now

lsimonsen (approved):
All good to go 🚀

PR #540, opened 2022-10-07
updating readme

CyrusYavariMacro:
Updating readme for extra explanations, also updating config file to …include development status.


setup.py
```diff
@@ -30,6 +30,7 @@
 Operating System :: Microsoft :: Windows
 Operating System :: POSIX
 Operating System :: MacOS
+Development Status :: 4 - Beta
```
lsimonsen, line 33: Can we set this up to automatically be updated from the ISRELEASE boolean?

lsimonsen (approved):

PR #556, opened 2022-10-10
Suggestion for PyPI version checks

lsimonsen:
Suggestion for checking if versions has been updated.

Suggestion of logic:

1. The version of the package needs to be greater or equal to the one on test.pypi.
2. For PyPI itself, it needs to always be greater (as the published package is immutable).

What do you think @Magnus167 ?


Magnus167 (approved):
Updated code reviewed ✔️

Magnus167 (approved):
Renamed file

PR #595, opened 2022-11-14
reestimation frequency historic vol

Magnus167:
This branch changes how the historic volatility function runs. 
Till now, the historic volatility was being calculated on every date in the input dataframe.


Magnus167:
In case of a daily reestimation freq., the program still uses `pandas.DataFrame.rolling`. However, in case of a monthly/weekly/quarterly reestimation freq., python loop is used to drive the calculation. There should be a way to use indexed addressing for the latter; which will dramatically increase execution speed (and readability).

rsueppel (approved):

PR #604, opened 2022-11-29
Feature/api refactor

Magnus167:


lsimonsen:
Suggestion for refactoring of overwall structure`

* download
  * dq_auth.py
  * dq_api.py
  * jpmaqs.py

and include a stub `dataquery` folder, where we can redirect from with a deprecation warning.

The main interface for users should be the callse `download.JPMaQSDownload` which has a `download` attribute. This way, if we need to change/add another data distribution source, we can easily add it as a key-word argument in the jpmaqs.py file.

macrosynergy/download/dq_api.py
```diff
@@ -0,0 +1,629 @@
...
+
+    with requests.request(
+        method=method,
+        url=url,
+        cert=cert,
+        headers=headers,
+        params=params,
+        **kwargs,
+    ) as r:
+        text: str = r.text
```
lsimonsen, line 67: It looks to me like we don't need the `text` variable?

Also, isn't `last_url` captured in `msg` (or we can add it to always be captured in msg?

macrosynergy/download/dq_api.py
```diff
@@ -0,0 +1,629 @@
...
+        assert isinstance(directory, str), dir_error
+
+        condition = os.path.exists(directory) and os.path.isfile(directory)
+        if not condition:
+            raise OSError(
+                f"The directory received, {directory}, does not contain the "
+                f"respective file, {file_type}."
+            )
+
+        return directory
```
lsimonsen, line 138: I think we need to be clearer in our language, as isn't technically what we are checking here whether the filename (and path of the file) is valid and exists.

macrosynergy/download/dq_api.py
```diff
@@ -0,0 +1,629 @@
...
+    def __init__(
+        self,
+        oauth: bool = False,
+        debug: bool = False,
+        concurrent: bool = True,
+        batch_size: int = 20,
+        **kwargs,
+    ):
+
+        self.proxy = kwargs.pop("proxy", None)
```
lsimonsen, line 293: Do we need to add your changes to the proxy of:
```python
self.proxy = kwargs.pop("proxy", kwargs.pop("proxies", None))
```

macrosynergy/download/jpmaqs.py
```diff
@@ -0,0 +1,478 @@
...
+        client_secret: Optional[str] = None,
+        debug: bool = False,
+        suppress_warning: bool = False,
+        **kwargs,
+    ):
+
+        self.debug = debug
+        self.suppress_warning = suppress_warning
+
+        proxy = kwargs.get("proxy", None)
```
lsimonsen, line 28: Similar, do we want to add the proxies option here as well?

lsimonsen (changes requested):
A few minor suggestions.

macrosynergy/dataquery/api.py
```diff
@@ -83,26 +85,26 @@ def check_connection(self) -> Tuple[bool, dict]:
         """Check connection (heartbeat) to DataQuery.
         """
         endpoint = "/services/heartbeat"
-        js: dict = self.access.get_dq_api_result(
+        js, success, msg = self.access.get_dq_api_result(
             url=self.access.base_url + endpoint,
             params={"data": "NO_REFERENCE_DATA"},
             proxy=self.proxy
         )
```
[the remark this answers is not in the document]
Magnus167: Since `check_connection()` should return a boolean, I think it makes sense to return`(results["code"]==200)`. Since the `results` dict is also returned, processes calling `check_connection()` have access to `results["code"]`.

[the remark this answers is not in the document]
Magnus167: Is it a good idea to rewrite 'check_connection()' such that it returns `Tuple[bool, dict]` with `(results["code"]==200)` & `results` when called as `check_connection(debug=True)`?
We could then set `check_connection()` to return only a boolean when called without args. This would mean keeping `debug=False` by default, or using `self.debug` from the Interface/JPMaQSDownload object.

macrosynergy/dataquery/api.py
```diff
@@ -147,25 +149,21 @@ def _fetch_threading(self, endpoint, params: dict, server_count: int = 5):
...
             try:
                 # The required fields will already be instantiated on the instance of the
                 # Class.
-                response: dict = self.access.get_dq_api_result(url=url, params=params, proxy=self.proxy)
+                response, status, msg  = self.access.get_dq_api_result(url=url, params=params, 
+                                                                    proxy=self.proxy)
             except ConnectionResetError:
                 counter += 1
                 time.sleep(0.05)
                 print(f"Server error: will retry. Attempt number: {counter}.")
```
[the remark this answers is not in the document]
Magnus167: Changed print statement to log statement. 
Breaking here would imply that the error tickers are not retried by the thread driver (`_request`). 
The thread driver breaks the code in case of an erroneous result, and re-runs the `error_tickers`. 
My concern with retrying individual threads is that it could result in too many hits to dataquery

macrosynergy/dataquery/api.py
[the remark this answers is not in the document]
Magnus167: The re-write is done, and has the assert statement as of now. However, I think it's problematic as it causes `check_connection()` to break (and thus the rest of the code) and not return False. 
So the only valid state for `check_connection()`'s return data would become `[True, {results}]`

macrosynergy/download/dq_api.py
```diff
@@ -0,0 +1,749 @@
...
+
+        valid_results_count = (
+            len(results) - len(invalid_expressions) - len(unavailable_expressions)
+        )
+        logger.warning(f"Invalid expressions: {', '.join(invalid_expressions)}")
+        logger.warning(f"Number of invalid expressions: {len(invalid_expressions)}")
+        logger.warning(f"Unavailable expressions: {', '.join(unavailable_expressions)}")
+        logger.warning(
+            f"Number of unavailable expressions: {len(unavailable_expressions)}"
+        )
```
Magnus167, line 730: @lsimonsen; Are these conditions valid checks for unavailable & invalid expressions?

macrosynergy/download/dq_api.py
```diff
@@ -0,0 +1,749 @@
...
+                    "access_token": js["access_token"],
+                    "expires_in": js["expires_in"],
+                }
+            except:
+                raise runtime_error
+            else:
+                if self._valid_token():
+                    return self._stored_token["access_token"]
+        else:
+            raise runtime_error
```
lsimonsen, line 256: Why `RuntimeError`?

macrosynergy/download/dq_api.py
```diff
@@ -0,0 +1,749 @@
...
+        i = 0
+        success = False
+        while (not success) and (i < 3):
+            js, success, self.last_url, msg = dq_request(
+                url=self.__token_url,
+                data=self.token_data,
+                method="post",
+                proxies=self.token_proxy,
+            )
+            i += 1
```
lsimonsen, line 240: Why are we trying to request multiple times?

lsimonsen: I wonder if we are better off using off the shelves libraries for the OAuth handling:

```python
from oauthlib.oauth2 import BackendApplicationClient
from requests_oauthlib import OAuth2Session

OAUTH_BASE_URL: str = "https://api-developer.jpmorgan.com/research/dataquery-authe/api/v2/"
OAUTH_TOKEN_URL: str = "https://authe.jpmchase.com/as/token.oauth2"
OAUTH_DQ_RESOURCE_ID: str = "JPMC:URI:RS-06785-DataQueryExternalApi-PROD"

class DQOAuth(object):
    def __init__(self, client_id: str, client_secret: str):
         self.client_id: str = client_id
         self.client_secret = client_secret

         self.client = BackendApplicationClient(client_id=client_id, aud=OAUTH_DQ_RESOURCE_ID)
         self.session = OAuth2Session(client=client)

    def _renew_token(self):
         self.session.fetch_token(
             token_url=OAUTH_TOKEN_URL,
             client_id="<client_id>",
            client_secret="<client_secret>"
          )

   def authorized(self):
        return self.session.authorized

    def request(url: str, method: = "GET"):
        if not self.authorized():
             self._renew_token()


        with self.session.request(url=url, method=method) as r:
             assert r.ok
             js = r.json()
        return js
```

macrosynergy/download/dq_api.py
```diff
@@ -0,0 +1,749 @@
+"""DataQuery Interface."""
```
lsimonsen, line 1: Why not name this file `dataquery.py` to make it simpler to read?

macrosynergy/download/dq_api.py
```diff
@@ -0,0 +1,749 @@
...
+
+    """
+
+    def __init__(
+        self,
+        username: str,
+        password: str,
+        crt: str = "api_macrosynergy_com.crt",
+        key: str = "api_macrosynergy_com.key",
+        base_url: str = CERT_BASE_URL,
```
lsimonsen, line 107: No proxies option?

macrosynergy/download/dq_api.py
Magnus167: The logic behind the multiple requests was to retry getting the token 3 times if and when it a token renew request fails. I've reverted to the original method (currently in use).

Magnus167: I'll look into using the library methods directly. Seems cleaner to use as well.

macrosynergy/download/dq_api.py
Magnus167: This was for the case when the process fails to get a new valid token. I had assumed it was the correct type of error to use; have reverted to the original method as of now

lsimonsen:
```python
class AuthenticationError(RuntimeError):
    """HTTP AuthenticationError"""
```

macrosynergy/download/exceptions.py
```diff
@@ -0,0 +1,7 @@
+""" Exception Classes for the download module. """
+
+
+class AuthenticationError(Exception):
+    """Raised when authentication fails."""
+
+    pass
```
lsimonsen, line 7: No need for the `pass` here.

tests/unit/download/test_api.py
```diff
@@ -0,0 +1,297 @@
...
+        with self.assertRaises(AssertionError):
+            dataquery.OAuth(client_id=b"test-id", client_secret="SECRET")
+
+    def test_invalid_dtype_client_secret(self):
+        with self.assertRaises(AssertionError):
+            dataquery.OAuth(client_id="test-id", client_secret=b"SECRET")
+
+    def test_valid_token(self):
+        oauth = dataquery.OAuth(client_id="test-id", client_secret="SECRET")
+        self.assertFalse(oauth._valid_token())
```
lsimonsen, line 36: Won't this make an actually call to DQ? And if so, shouldn't it be an integration test instead of unit test?

Magnus167: OAuth._valid_token does not make a call to DQ; it only checks the state of the stored token (if there is one. if not, it returns false)

macrosynergy/download/dataquery.py
```diff
@@ -0,0 +1,743 @@
...
+        url = self.access.base_url + endpoint
+        select = "instruments"
+        response = {}
+        results = []
+        counter = 0
+        invalid_responses = 0
+        while (
+            (not (select in response.keys()))
+            and (counter <= server_count)
+            and (invalid_responses <= server_count)
```
lsimonsen, line 402: Wouldn't we arbitrarily break after 5 retries now as default? Even if the next link has more left?

macrosynergy/download/dataquery.py
```diff
@@ -0,0 +1,743 @@
...
+
+        return <dict>: single dictionary containing all the requested Tickers and their
+            respective time-series over the defined dates.
+        """
+
+        if delay > 0.9999:
+            error_delay = "Issue with DataQuery - requests should not be throttled."
+            raise RuntimeError(error_delay)
+
+        no_tickers = len(tickers)
```
lsimonsen, line 532: We should rename tickers to `expressions` to follow the DataQuery convention in this part.

macrosynergy/download/dataquery.py
```diff
@@ -0,0 +1,743 @@
...
+                    error_tickers = rec_error_tickers
+                    error_messages.extend(rec_error_messages)
+                    final_output.extend(rec_final_output)
+                    if not error_tickers:
+                        recursive_call = False
+                    elif count > 5:
+                        recursive_call = False
+                        logger.warning(
+                            f"Error requesting tickers: {', '.join(error_tickers)}. No longer retrying."
+                        )
```
lsimonsen, line 646: Shouldn't this be further up? in first `error_tickers`?

macrosynergy/download/jpmaqs.py
```diff
@@ -0,0 +1,536 @@
...
+import sys
+
+# check if parent logger exists. else, create it.
+if not logging.getLogger().hasHandlers():
+    logging.basicConfig(
+        stream=sys.stderr,
+        level=logging.WARNING,
+        format="%(asctime)s - %(levelname)s - %(funcName)s - %(message)s",
+        datefmt="%m/%d/%Y %I:%M:%S %p",
+    )
```
lsimonsen, line 19: Never modify loggers inside code :)

lsimonsen: Even if it means users won't see the output.

macrosynergy/download/dataquery.py
Magnus167: Yes; that's true. 
I think setting `server_count=10` and testing `(counter+invalid_responses)<=server_count` before breaking might be a better idea.
PS: the variables after refactor are 
`server_count`→`max_retries`
`counter`→`conxn_errors`

macrosynergy/download/dataquery.py
```diff
@@ -0,0 +1,743 @@
...
+            f" with message: {msg}"
+            f" and response: {js}"
+        )
+        if msg["status_code"] == 401:
+            logger.error(
+                f"Invalid credentials."
+                f"Request failed for URL: {last_url}"
+                f" with message: {msg}"
+                f" and response: {js}"
+            )
```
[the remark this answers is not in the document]
lsimonsen:
```python
logger.error(
    "Invalid credentials. Request failed for URL: %s"
    " with message: %s and response: %s",
    last_url, msg, js
)
```

macrosynergy/download/jpmaqs.py
Magnus167: Removed

lsimonsen (approved):

PR #614, opened 2022-12-08
adding buisness days dif to be called from the package

CyrusYavariMacro:


Magnus167 (approved):
business_day_dif function added to management module __init__

PR #626, opened 2023-01-09
API changes for ChunkedEncodingError and recording DQ Error messages

Magnus167:


macrosynergy/download/dataquery.py
```diff
@@ -743,6 +743,8 @@ def get_ts_expression(
         c_delay = self.delay_compute(len(expressions))
         results = None
 
+        print(datetime.utcnow().strftime("%Y-%m-%d %H:%M:%S.%f")[:-3] + " UTC")
```
lsimonsen, line 746: Alternative: 
```python
print(datetime.utcnow().isoformat())
```

lsimonsen (approved):

macrosynergy/download/dataquery.py
Magnus167: Changed to `print(datetime.utcnow().isoformat(), " UTC")` to avoid confusion

lsimonsen (approved):

PR #674, opened 2023-02-22
Feature : Slip arg in CategoryRelations

Magnus167:


rsueppel (changes requested):
There seem to be two misunderstandings:

1.  Slippage is another source of delay in the application of information and, hence, it is principally additive to the lag.
2. Slippage is a delay that is being applied to the original frequency (always days in our case), not the potentially downsampled frequency. It reflects the time that is required for implementing a position and, by implication, the target values that are applied. This cannot be implemented with the standard lag because [a] that lag delays the feature (while we need to bring forward the target) and [b] the lag is applied in the downsampled frequency.

What must be done is that slip is used up-front in the code to bring forward the target by the given number of days. Do not delay the feature, since the two operations are not equivalent.

rsueppel (approved):
In pandas, a .shift(k) with k>0 means that values arrive later on the time axis. Positive slip means a negative shift of target, i.e. target values arrive earlier, by virtue of lines 164-168. This means target values are associated with older signal values. After downsampling the effective downsampled period returns drop early days and carry over to early days of the next period.

PR #686, opened 2023-03-02
Feature : View timelines, reset index to standardize input DF

### LINKED ISSUE #685: view_timelines: resent index of input dataframe
In order to avoid "cannot reindex on an axis with duplicate labels" errors, apply reset_index(drop=True) to the input df.

Magnus167:
Closes #685.

Added a simple code block to check if the input DF has the necessary columns of a standard DF.
If not, `df.reset_index` is attempted; if that works, good - continue. Else raises exception.


rsueppel (changes requested):
Why do you only reset the index if the columns do not include the expected standard?

rsueppel (approved):

PR #688, opened 2023-03-04
Bugfix : DQ Last Date with default args

Magnus167:
For some reason, the Dataquery endpoints fail to return published data for the current day. This PR "forces" the download client to request "future dates" (today + 2bdays) to ensure that any published data for today is downloaded.

### DIFF
--- a/macrosynergy/download/jpmaqs.py
+++ b/macrosynergy/download/jpmaqs.py
@@ -625,7 +625,8 @@ class JPMaQSDownload(object):
             else:
                 df = df.sort_values(["cid", "xcat", "real_date"]).reset_index(drop=True)
                 
-                
+                # remove any entries for after "today""
+                df = df[df["real_date"].isin(pd.date_range(start=start_date, end=end_date))]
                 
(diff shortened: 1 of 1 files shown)


macrosynergy/download/jpmaqs.py
```diff
@@ -486,7 +487,14 @@ def download(
                 "mop_lag",
                 "grading",
             ], f"Incorrect metric passed: {metric}."
+            
+        assert isinstance(start_date, str), "Start date must be a string in the format YYYY-MM-DD"
+        if end_date is not None:
+            assert isinstance(end_date, str), "End date must be a string in the format YYYY-MM-DD"
+        else:
+            end_date = (datetime.datetime.today() + pd.offsets.BusinessDay(2)).strftime("%Y-%m-%d")
```
lsimonsen, line 495: Ah, this won't work Palash, unfortunately DQ doesn't take a ISO format date, but rather "%y%m%d" as the string - if you check the docs.

lsimonsen (changes requested):

PR #706, opened 2023-03-31
Feature : make_relative_value basket 

### LINKED ISSUE #683: Enhance flexibility of basket calculation in `make_relative_value`
So far the `basket` argument can only take cross-sections for basket formation that are also in `cids`, i.e. a subset. 

It should also allow other cross-sections in the data frame.

Magnus167:
closes #683


rsueppel (changes requested):
Palash:

There is no need for an additional parameter (cids_all). All that needs to change is that the list in `basket` can include cross-section identifies that are not in `cids`. Put simply, `cids` determines the list of cross-section for which relative values are calculated, while `basket` determines the cross-sections that constitute the basket. Cheers

rsueppel (dismissed):

rsueppel (approved):

PR #774, opened 2023-05-03
Bugfix : `CategoryRelations.apply_slip`

Magnus167:


rsueppel (changes requested):
Consider merging #781 instead

rsueppel (approved):

PR #775, opened 2023-05-03
Feature : Code format workflow

Magnus167:


lsimonsen (changes requested):
I don't think this will work, as currently you can push to main, test and develop - as it requires code reviews.

Also, we will need to add the user to the repository (I think) or is that handled by the access token?

rsueppel (approved):

PR #794, opened 2023-05-12
Bugfix: `test_package.py`, Failing on Windows systems

Magnus167:
Fails on a Windows system where the user may have a space such as `C:/User Name/script.py`

### DIFF
--- a/tests/unit/test_package.py
+++ b/tests/unit/test_package.py
@@ -9,7 +9,7 @@ class TestPackage(unittest.TestCase):
         cmd = ("import macrosynergy as msy; "
                "import sys; ")
 
-        cmd = sys.executable + ' -c "' + cmd + '"'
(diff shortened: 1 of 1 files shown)


tests/unit/test_package.py
```diff
@@ -9,15 +9,15 @@ def test_lazy_imports(self):
         cmd = ("import macrosynergy as msy; "
                "import sys; ")
 
-        cmd = sys.executable + ' -c "' + cmd + '"'
+        cmd = f'"{sys.executable}" -c "' + cmd + '"'
         p = subprocess.Popen(cmd, shell=True, close_fds=True)
```
lsimonsen, line 13: Why not run it as:
```python
p = subprocess.Popen([sys.executable, "-c", cmd], shell=True, close_fds=True)
```

lsimonsen (approved):

PR #880, opened 2023-07-04
Bugfix: Updating docstring for `linear_composite()`

Magnus167:


rsueppel (changes requested):
Add to signs docstring: "Default is positive signs only".

rsueppel (approved):

PR #891, opened 2023-07-07
WIP : Refactoring `linear_composite()`

Magnus167:


macrosynergy/panel/linear_composite.py
```diff
@@ -1,173 +1,191 @@
...
+    if normalize_weights:
+        adj_weights_wide = weights_df[~nan_mask].div(
+            weights_df[~nan_mask].abs().sum(axis=1), axis=0
+        )
+        adj_weights_wide[nan_mask] = np.NaN
+
+        assert np.allclose(
+            adj_weights_wide.abs().sum(axis=1), 1
+        ), "Weights do not sum to 1. Normalization failed."
+
```
rsueppel, line 39: lines 36-38 would only activate if rows are all NaNs, which should be checked beforehand on nan_mask and a more specific message should be given.

macrosynergy/panel/linear_composite.py
```diff
@@ -1,173 +1,191 @@
...
+    # Multiply the weights by the target data
+    out_df = data_df * adj_weights_wide
+
```
rsueppel, line 42: Line 41 would throw an error if normalize_weights is False

macrosynergy/panel/linear_composite.py
```diff
@@ -1,173 +1,191 @@
...
+    # Remove periods with missing data (if requested)
+    if complete:
+        out_df[nan_mask.any(axis=1)] = np.NaN
     else:
```
rsueppel, line 46: lines 46-47 look redundant as missing rows imply that weights don't add to 1.

macrosynergy/panel/linear_composite.py
```diff
@@ -1,173 +1,191 @@
...
-        out_df[mask.all(axis=1)] = np.NaN
+        out_df[nan_mask.all(axis=1)] = np.NaN
 
+    # Sum across the columns
+    out_df = out_df.sum(axis=1)
```
rsueppel, line 50: line 50 calculates incorrect values
"By default, the sum of an empty or all-NA Series is 0 . This can be controlled with the min_count parameter."

macrosynergy/panel/linear_composite.py
Magnus167: That's correct. This is intended as a sanity check. Will update the message

macrosynergy/panel/linear_composite.py
Magnus167: fixed.

macrosynergy/panel/linear_composite.py
Magnus167: Have changed it out to remove all rows with all nans, before we sum the columns

macrosynergy/panel/linear_composite.py
Magnus167: Now resolved; any rows with all nans are simply dropped

rsueppel (approved):

PR #956, opened 2023-09-06
Feature: Added ability to pass directives in docstrings

Magnus167:
Currently functional directives-

```
::docs::function_name::sort_first::
::docs::function_name::sort_last::
```

As of now, all directives within a module are parsed together. i.e. it wouldn't matter if you were addressing some other function in a functions docstring.
For this reason, it is recommended to list these directives in the module docstring.
Future directives can include highlighting, more sorting, and even appending examples


macrosynergy/download/dataquery.py
```diff
@@ -541,7 +543,7 @@ def validate_download_args(
     return True
 
 
-def get_unavailable_expressions(
+def _get_unavailable_expressions(
```
Magnus167, line 546: eg. alternatively, this could have been changed by adding: 
```
::docs::get_unavailable_expressions::sort_last
```

Issue #967, opened 2023-09-11
panel_calculator cannot handle single series


rsueppel:
The panel_calulator function throws an error if a global series, using the prefix "i", i s alone on the right hand of the formula. 

![Image](https://user-images.githubusercontent.com/27678335/266994236-e7f15aa6-654c-4af0-b7f7-c72089e77294.png)

This is a defect because often we just want to assign one series to all series of a panel.
One can avoid the error by putting an irrelevant category to the right hand side of the equation, but it just looks awful and confusing:

![Image](https://user-images.githubusercontent.com/27678335/266994566-81e0ed70-af28-41c1-8edd-bd643b953da8.png)

PR #968, opened 2023-09-11
Bugfix: `panel_calculator()`, error with incomplete CID-XCAT intersection

Magnus167:
Closes issue #967

### DIFF
--- a/macrosynergy/panel/panel_calculator.py
+++ b/macrosynergy/panel/panel_calculator.py
@@ -156,11 +156,17 @@ def panel_calculator(df: pd.DataFrame, calcs: List[str] = None,
     single_xcats = [x[5:] for x in singles_used]
     all_xcats_used = xcats_used + single_xcats
 
-    new_xcats = list(ops.keys())
-    old_xcats_used = set(all_xcats_used) - set(new_xcats)
-    old_xcats_used = list(old_xcats_used)
-    missing = sorted(set(old_xcats_used) - set(df['xcat'].unique()))
-    assert len(missing) == 0, f"Missing categories: {missing}."
+    new_xcats: List[str] = list(ops.keys())
+    old_xcats_used: List[str] = list(set(all_xcats_used) - set(new_xcats))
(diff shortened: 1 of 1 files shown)


macrosynergy/panel/panel_calculator.py
```diff
@@ -156,11 +156,17 @@ def panel_calculator(df: pd.DataFrame, calcs: List[str] = None,
...
-    assert len(missing) == 0, f"Missing categories: {missing}."
+    new_xcats: List[str] = list(ops.keys())
+    old_xcats_used: List[str] = list(set(all_xcats_used) - set(new_xcats))
+    missing: List[str] = sorted(set(old_xcats_used) - set(df['xcat'].unique()))
+    if len(missing) > 0:
+        raise ValueError(f"Missing categories: {missing}.")
+
+    if len(singles_used) > 0:
+        if new_xcats == all_xcats_used:
+            old_xcats_used: List[str] = new_xcats
```
rsueppel, line 167: It seems that the intention is to prevent an undue data frame reduction in cases where only new xcats are used. 
However, why would be ever want to drop the panel dimensions as defined by cids? Why would that be necessary?
Even if we just pass the string: "XCAT1 = 1" for cids = cidx, it should still calculate a panel of 1s.

Magnus167: A few lines above 165, 
```python
old_xcats_used = list(set(all_xcats_used) - set(new_xcats))
... 
# after line 167
reduce_df(xcats=old_xcats_used,...)
```

In case where an XCAT does not exist for a particular CID, it will be in `new_xcats` as well as `all_xcats_used`.
With context to the example in #967:
```python
formula = "RIR_NSA = iEUR_RIR_NSA"
```
implying 
`all_xcats_used == new_xcats`; 

and therefore 
`old_xcats_used == []`

Re. enhancement for allowing numerical assignment:
As of now, the example `"XCAT = 1"` will not work with the panel calculator. There are multiple steps at which an operation like this would fail. This would be better solved as an enhancement

rsueppel (approved):

PR #1030, opened 2023-10-03
Bugfix: `NaivePNL`, daily rebal and slip

Magnus167:


macrosynergy/pnl/naive_pnl.py
```diff
@@ -195,12 +195,17 @@ def __make_signal__(
...
     @classmethod
-    def rebalancing(cls, dfw: pd.DataFrame, rebal_freq: str = "daily", rebal_slip=0):
+    def rebalancing(
+        cls, dfw: pd.DataFrame, rebal_freq: str = "daily", rebal_slip: int = 0
+    ) -> np.ndarray:
         """
         The signals are calculated daily and for each individual cross-section defined in
         the panel. However, re-balancing a position can occur more infrequently than
         daily. Therefore, produce the re-balancing values according to the more
-        infrequent timeline (weekly or monthly).
```
rsueppel, line 203: Add quarterly rebalancing as an option while you are at it.

The docstring on daily rebalancing should be: The re-balancing can also be done on a
        daily basis. Since this incurs potentially high transaction costs this is not the standard
       for naive PnLs, however.

PR #1041, opened 2023-10-09
Feature: extend `correl_matrix` to calculate correlation between two sets

emhbrine:
This PR allows the `xcats` or `cids` parameters of the `correl_matrix` function to take a list containing two lists. In this case, the correlation will be calculated and visualized between these two sets. 

If both sets contain a single category, such as `xcats = [['XR'], ['CRY']]`, correlation will be visualized between the cross-sections of these categories. If either set contains more than one category, correlation is visualized between categories.

cross_section_corr
cross_category_corr

### DIFF
--- a/macrosynergy/panel/view_correlations.py
+++ b/macrosynergy/panel/view_correlations.py
@@ -162,6 +162,7 @@ def correl_matrix(
     xlabel = ""
     ylabel = ""
 
+    # If more than one set of xcats or cids have been supplied.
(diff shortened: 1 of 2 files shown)


Magnus167 (dismissed):
approved @emhbrine's changes

Magnus167 (approved):
Approved @emhbrine 's

Magnus167 (approved):

PR #1047, opened 2023-10-09
Feature: Caches dependencies when building

sandresen1:
On every build, caching should take place to store the pip dependencies so they don't need to be reinstalled each time upon building.


sandresen1:
Caching seems to work fine and will only install a package if specified in the toml file. But the time taken to retrieve from cache seems to be the same as downloading :(

Magnus167 (dismissed):

emhbrine (approved):

PR #1110, opened 2023-10-23
small edits to documentation

rsueppel:


rsueppel:
Reminders:

- [ ] implement freqs as pre-selection of frequencies inside the class
- [ ] document manipulate_df and all all methods
- [ ] carefully check all output stats with unit test that calculate those same stats independently

sandresen1 (approved):

PR #1124, opened 2023-10-25
Feature: Prevent Build from succeeding if sys.path.append(os.getcwd()) is found

sandresen1:


Magnus167 (approved):
That's a painful 40 commits

PR #1148, opened 2023-10-26
Feature: Adding `granger_causality_test()`

Magnus167:
Added a function that runs a Granger causality test with the following parameters and dosctrings.
```
Run Granger Causality Test on a standardized quantamental dataframe.
Since the Graner Causality Test is a pairwise test, only two tickers can be specified.
Specify either `tickers` or `cids` & `xcats`. When specifying `cids` & `xcats`, the
user may input one `cid` and two `xcats`; or two `cids` and one `xcat` to yield two
tickers. The function forms the list of tickers from the `cids` and `xcats` such that
the order of the formed tickers is preserved. The order of the tickers is important
as the first ticker is the one that is tested to Granger cause the second ticker.
The function tests whether the first ticker Granger causes the second ticker.

:param <pd.DataFrame> df: A standardized quantamental dataframe.
:param <List[str]> tickers: A list of tickers to run the test on. A maximum of two
    tickers can be specified.
:param <Union[str, List[str]]> cids: One or two cids to run the test on. If two cids
    are specified, then only one xcat can be specified. If one cid is specified, then
    two xcats can be specified.
:param <Union[str, List[str]]> xcats: One or two xcats to run the test on. If two
    xcats are specified, then only one cid can be specified. If one xcat is specified,
    then two cids can be specified.
:param <Union[int, List[int]]> max_lag: If `max_lag` is an integer, then the function
    computes the test for all lags up to `max_lag`. If `max_lag` is a list of integers,
    then the function computes the test only for lags specified in the list.
:param <bool> add_constant: Whether to add a constant to the regression.
:param <str> freq: The frequency to downsample the data to. Must be one of "D", "W",
    "M", "Q", "A". Default is "M".
:param <str> agg: The aggregation method to use when downsampling the data. Must be
    one of "mean" (default), "median", "min", "max", "first" or "last".
:param <str> start: The start date of the data. Must be a valid ISO date. If not
    specified, the earliest date in `df` is used.
:param <str> end: The end date of the data. Must be a valid ISO date. If not
    specified, the latest date in `df` is used.
:param <str> metric: The metric to run the test on. Must be a column in `df`.
    Default is "value".
```


rsueppel (commented):
This should be an experimental internal-only function at this stage. I added an issue for a proper package class but it does not have top priority right now.

sandresen1 (approved):

sandresen1 (approved):

PR #1173, opened 2023-11-01
Feature: Add more unit tests to Signal Return Relations

sandresen1:


macrosynergy/signal/signal_return_relations.py
```diff
@@ -361,6 +364,9 @@ def correlation_bars(
         plt.legend(loc=legend_pos)
         plt.show()
 
+        if testing:
```
emhbrine, line 367: It's best to avoid introducing code used only in tests

emhbrine (approved):

rsueppel (approved):

PR #1178, opened 2023-11-03
Bugfix: Clips expected zn scores to prevent occasional false failure on modify signals test

sandresen1:


Magnus167 (approved):

Magnus167:
Problem was: Test did not correctly simulate the expected behaviour of `thresh` param

PR #1235, opened 2023-11-20
Feature: Deprecating singular arguments in Signal Return Relations

sandresen1:


macrosynergy/signal/signal_return_relations.py
```diff
@@ -62,26 +57,86 @@ class SignalReturnRelations():
...
+        freqs: Union[str, List[str]] = "M",
+        agg_sigs: Union[str, List[str]] = "last",
+        freq: str = None,
+        agg_sig: str = None,
         fwin: int = 1,
         slip: int = 0,
     ):
+        if rets is None:
+            if ret is None:
+                raise ValueError("Target return must be defined.")
```
Magnus167, line 103: Can you add tests to check for the value errors?

Magnus167 (approved):

PR #1365, opened 2024-01-04
Feature: Sign- and time-weighted regression models

rushilg99:


macrosynergy/learning/predictors.py
```diff
@@ -1,16 +1,391 @@
...
+        :param <RegressorMixin> regressor: Scikit-learn regressor class to be used as the
+            underlying model for which a sign-weighted fit is to be performed.
+
+        NOTE: This class is specifically tailored for compatibility with any scikit-learn
+        regressor that supports a 'sample_weight' parameter in its 'fit' method. Each
+        training sample receives a weight inversely proportional to the frequency of its
+        label's sign in the training set. Thus, the model is encouraged to learn equally
+        from both positive and negative return samples.
+        """
+        if not callable(regressor) or not inspect.isclass(regressor):
```
emhbrine, line 46: Can this and the next check be replaced by `if not isinstance(regressor, RegressorMixin) or not isinstance(regressor, BaseEstimator)`?

macrosynergy/learning/predictors.py
```diff
@@ -1,16 +1,391 @@
...
+        scikit-learn regressor that supports a 'sample_weight' parameter in its 'fit'
+        method. Each training sample receives a weight based on the date attributed to
+        the sample. The weights are calculated as an exponentially decaying function of
+        the date, with the half-life of the decay specified by the user.
+        Consequently, during the model training process, more recent samples
+        are emphasized due to the higher weight they receive. By weighting based on
+        recency, the model is encouraged to prioritise newer information.
+        """
+        # Checks
+        if not callable(regressor) or not inspect.isclass(regressor):
```
emhbrine, line 191: Also here

macrosynergy/learning/predictors.py
rushilg99: No. Otherwise, LinearRegression and LinearRegression() would both pass but the second is not valid here -> for this class, we want all hyperparameters for LinearRegression (or another estimator) to be passed separately to allow for hyperparameter optimisation of those whilst using a weighted loss function.

emhbrine: Ah I see!

rsueppel (approved):

PR #1376, opened 2024-01-08
Feature: AUC and Macrosynergy panel test p-value statistics added to signal return relations

sandresen1:


macrosynergy/signal/signal_return_relations.py
```diff
@@ -18,6 +22,8 @@
 )
 import macrosynergy.visuals as msv
 
+np.seterr(divide="ignore")
```
emhbrine, line 25: I'd suggest we use `with np.errstate(divide='ignore'):` rather than disabling it globally

Magnus167 (approved):

PR #1442, opened 2024-02-04
Feature: Optimize dataframe conversion in `JPMaQSDownload()`

Magnus167:
- Reduces API_DELAY_PARAM to 0.2 (200ms) from 0.3
- Removes checks for duplicates. as DQ specifies that only unique date-value pairs will make it into a timeseries. Also removes the test for this
- Uses a new method for converting JSONs to DFs
- Adds progress notifications during dataframe conversion


macrosynergy/download/jpmaqs.py
```diff
@@ -372,42 +370,30 @@ def time_series_to_df(
...
+            ),
+            list(
+                map(
+                    lambda metricx: pd.concat(dfs_dict[metricx], ignore_index=True),
+                    dfs_dict,
                 )
+            ),
+        )
 
-            raise InvalidDataframeError(err_str)
```
Magnus167, line 403: Should we drop this check @emhbrine , @sandresen1 ?

sandresen1 (approved):

PR #1472, opened 2024-02-13
Feature: Signal Optimizer explainability

rushilg99:


rushilg99:
TODO: make sure custom regressors still store coefficients and intercepts

rushilg99:
TODO:

- [ ] Unit test for the case in the worker function where no model was selected
- [ ] Unit testing for a pipeline and not an indivisual estimator (particularly with feature selection)
- [ ] Unit test for the case where sklearn returns an invalid coefficient vector (if possible)
- [ ] Unit test for the case where sklearn returns an invalid intercept format
- [ ] Unit test for the case where multiple models are selected
- [ ] Make sure custom regressors store coefficients and intercepts

emhbrine (approved):

PR #1473, opened 2024-02-13
Feature:  View Timelines

sandresen1:
Optimise view timelines


macrosynergy/visuals/timelines.py
```diff
@@ -173,6 +163,7 @@ def timelines(
 
     cross_mean_series: Optional[str] = f"mean_{xcats[0]}" if cs_mean else None
     if cs_mean:
+        df = df.copy()
```
emhbrine, line 166: Just wondering why we need to copy it here but not otherwise?

emhbrine (approved):

PR #1517, opened 2024-03-01
Bugfix: Add check for empty dataframe in historic_vol function

Magnus167:


macrosynergy/panel/historic_vol.py
```diff
@@ -272,7 +277,7 @@ def single_calc(
 
         out_dfs.append(df_out.loc[sel_bools & sel_dts])
 
-    return standardise_dataframe(pd.concat(out_dfs))
+    return standardise_dataframe(pd.concat(out_dfs)).dropna().reset_index(drop=True)
```
lsimonsen, line 280: Palash, where is the N/A's coming from? We should make sure the is no gaps in the timeseries (only at the start can there be N/A's). Also, we should add logging and have a logging.info statement telling us what we are dropping.

PR #1522, opened 2024-03-04
ProxyPNL, `nan_tolerance` fix

Magnus167:


macrosynergy/pnl/historic_portfolio_volatility.py
```diff
@@ -129,15 +129,20 @@ def _estimate_variance_covariance(
 
     cov_matr = np.zeros((len(piv_ret.columns), len(piv_ret.columns)))
 
+    nan_ratio = lambda x: np.isnan(x).sum() + (x == 0).sum() / len(x)
+    apply_mask = lambda x: x if nan_ratio(x) <= nan_tolerance else np.nan * x
+
     for iB, cB in enumerate(piv_ret.columns):
         for iA, cA in enumerate(piv_ret.columns[: iB + 1]):
+            rA: np.ndarray = apply_mask(piv_ret[cA].values)
```
lsimonsen, line 137: Sorry for chase you on this one again @Magnus167, but [PEP8](https://peps.python.org/pep-0008/) requires lower case as the Python convention.

lsimonsen: Also, we need to apply the mask consistently for each, so they keep the same length between the two vectors `rA` and `rB` - especially as we drop the index when we make them into numpy arrays (`.values`).

lsimonsen (changes requested):

macrosynergy/pnl/historic_portfolio_volatility.py
```diff
@@ -120,31 +69,132 @@ def _estimate_variance_covariance(
...
+    :param <pd.DatetimeIndex> trigger_indices: the DateTimeIndex of the trigger dates.
+    :param <int> lback_periods: the number of periods to use for filling NaNs.
+    :param <float> nan_tolerance: the maximum ratio of NaNs to non-NaNs in a lookback
+        window, if exceeded the resulting volatility is set to NaN. Default is 0.25.
+    """
+
+    for trigger_date in trigger_indices:
+        dtidx = pivot_df.index[pivot_df.index <= trigger_date][-lback_periods:]
+        col_mask = (pivot_df.loc[dtidx].apply(_nan_ratio) > nan_tolerance).values
+        pivot_df.loc[dtidx, col_mask] = np.nan
```
lsimonsen, line 124: Unless I am mistaken this won't work. We need to impose the max nan tolerance per trigger date.

As an example:
```python
batches: List[Tuple[datetime.date, pd.DataFrame]] = []
for trigger_date in trigger_indices:
    df = pivot_df.loc[pivot_df.index <= trigger_date, :].iloc[-lback_periods:]
    col_mask = (df.apply(_nan_ratio, axis=0) > nan_tolerance).values
    df.loc[:, col_mask] = np.nan
    batches.append((trigger_date, df))
```
Obviously this then requires us to modify the logic later down. Alternatively we need to impose the N/A logic at a later stage.

lsimonsen (changes requested):

lsimonsen (changes requested):
A few questions in terms of use and functionality.

macrosynergy/pnl/historic_portfolio_volatility.py
```diff
@@ -38,113 +35,156 @@ def expo_weights_arr(lback_periods: int, half_life: int, *args, **kwargs) -> np.
...
-        rss = (x - x.mean()) * (y - y.mean())
-    else:
-        rss = (x - x.mean()) ** 2
+    # drop NaNs
+    x = x[~mask]
+    y = y[~mask]
 
-    return w.T.dot(rss)
+    # if dropping nans leaves less than lback_periods, fill with NaNs and return
+    if any(len(_) < lback_periods for _ in [x, y]):
```
lsimonsen, line 77: @Magnus167 in my books the use of `_` is a misnomer. Unless I am mistaken standard practise it to use `_` for unused variables?

Magnus167: It's used for a temporary variable; typically used when you want need a variable to pass/store/show a value, but don't need the variable afterwards.

macrosynergy/pnl/historic_portfolio_volatility.py
```diff
@@ -38,113 +35,156 @@ def expo_weights_arr(lback_periods: int, half_life: int, *args, **kwargs) -> np.
...
+                cov_matr[i_b, i_a] = np.nan
+                continue
+
+            w = weights_func(
+                lback_periods=min(lback_periods, len(x)),
+                half_life=min(half_life, len(x)),
+            )  # inherently fast, and cached anyway
+
+            est_vol = _weighted_covariance(x=x, y=y, w=w)
+            cov_matr[i_a, i_b] = est_vol
```
emhbrine, line 123: Should be named `cov_mat`

macrosynergy/pnl/historic_portfolio_volatility.py
```diff
@@ -457,5 +487,22 @@ def historic_portfolio_vol(
...
+
+    df_copy_vol: pd.DataFrame = historic_portfolio_vol(
+        df=df_copy,
+        sname="STRAT",
+        fids=fids,
+        est_freq="m",
+        lback_periods=15,
+        lback_meth="ma",
+        half_life=11,
+        rstring="XR",
```
emhbrine, line 499: `rstring` and `sname` should be renamed for clarity

macrosynergy/pnl/historic_portfolio_volatility.py
emhbrine: No mistake, @lsimonsen.
The underscore signifies that the variable should be ignored because it is not used in the loop.
It's also sometimes used in cases where a function might return multiple values, but some are to be ignored:
```
x, _ = func()
```

lsimonsen: I think I got it - the above `x, _ = func()` was what I was expecting (it never gets used). However, the point from you two, is that `_` can also be used inside the conditional statement `if statement`, when it doesn't get used afterwards/outside that evaluation.

lsimonsen: Thank you both @emhbrine and @Magnus167

macrosynergy/pnl/historic_portfolio_volatility.py
```diff
@@ -38,113 +35,156 @@ def expo_weights_arr(lback_periods: int, half_life: int, *args, **kwargs) -> np.
...
     assert np.all((cov_matr.T == cov_matr) ^ np.isnan(cov_matr))
 
     return pd.DataFrame(cov_matr, index=piv_ret.columns, columns=piv_ret.columns)
 
 
+def _calculate_portfolio_volatility(
+    trigger_indices: pd.DatetimeIndex,
+    pivot_returns: pd.DataFrame,
+    pivot_signals: pd.DataFrame,
+    lback_periods: int,
```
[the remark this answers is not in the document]
lsimonsen: I couldn't agree more. Logical and readable names is essential for clarity and ease of reading. However, as you point out Eric - it is a broader issue.

macrosynergy/pnl/historic_portfolio_volatility.py
lsimonsen: So `return_string`, and  `signal_name` essentially. Makes sense to me.

macrosynergy/pnl/historic_portfolio_volatility.py
```diff
@@ -38,113 +35,156 @@ def expo_weights_arr(lback_periods: int, half_life: int, *args, **kwargs) -> np.
...
+    """
+
+    return_values = []
+
+    for trigger_date in trigger_indices:
+        td = trigger_date
+        lbx = -1 * int(np.ceil(lback_periods * (1 + nan_tolerance)))
+        # account for possible NA drops and create a bigger slice
+        piv_ret = pivot_returns.loc[pivot_returns.index <= td].iloc[lbx:]
+        vcv: pd.DataFrame = _estimate_variance_covariance(
```
lsimonsen, line 170: Maybe add a comment that we are estimating the variance-covariance matrix? Or is this self-evident?

macrosynergy/pnl/historic_portfolio_volatility.py
```diff
@@ -38,113 +35,156 @@ def expo_weights_arr(lback_periods: int, half_life: int, *args, **kwargs) -> np.
...
+            piv_ret=piv_ret,
+            lback_periods=lback_periods,
+            *args,
+            **kwargs,
+        )
+
+        # TODO - NaN handing for signals?
+
+        piv_sig: pd.DataFrame = pivot_signals.loc[td, :]
```
lsimonsen, line 179: It isn't really a `pivot` (`piv`) but rather an array/series. Maybe rename to `signal: pd.Series = pivot_signals.loc[td, :]`?

lsimonsen (changes requested):

macrosynergy/pnl/historic_portfolio_volatility.py
```diff
@@ -38,113 +35,156 @@ def expo_weights_arr(lback_periods: int, half_life: int, *args, **kwargs) -> np.
...
+    return w.T.dot(rss)
 
 
-def weighted_covariance(
-    x: np.ndarray, y: np.ndarray, w: np.ndarray = None, remove_zeros: bool = True
-):
-    """
-    Estimate covariance between two series after applying weights.
+def _nan_ratio(x, remove_zeros: bool = True) -> float:
+    return (sum(np.isnan(x)) + (sum(x == 0) if remove_zeros else 0)) / len(x)
```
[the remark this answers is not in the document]
Magnus167: Will be removed soon

lsimonsen (approved):

PR #1527, opened 2024-03-05
`ProxyPNL`, adding `estimate_release_frequency` function

Magnus167:


macrosynergy/management/constants.py
```diff
@@ -12,18 +39,34 @@
...
+    "D": 1,
+    "W": 5,
+    "M": 24,
+    "Q": 64,
+    "A": 252,
+    "B": 1,
+    "W-FRI": 5,
+    "BM": 24,
+    "BQ": 64,
+    "BA": 252,
```
lsimonsen, line 53: What are these frequencies? Are they mapped to standard Pandas frequencies?

macrosynergy/pnl/notional_positions.py
```diff
@@ -111,7 +111,7 @@ def notional_positions(
     dollar_per_signal: Numeric = 1.0,
     leverage: Optional[Numeric] = None,
     vol_target: Optional[Numeric] = None,
-    rebal_freq: str = "m",
+    rebal_freq: str = "M",
```
lsimonsen, line 114: Why capital "M" - standard JPMaQS/Macrosynergy notation would be lower case `m` - we need to make it consistent then across the Macrosynergy package.

macrosynergy/management/utils/df_utils.py
```diff
@@ -300,6 +300,65 @@ def downsample_df_on_real_date(
...
+        if not "real_date" in df.columns:
+            raise ValueError("DataFrame must have a 'real_date' index")
+        df = df.set_index("real_date")
+
+    rdf = df.resample(freq).agg(agg, numeric_only=True)
+    rdf.index.name = "real_date"
+    if not (ffill or bfill):
+        return rdf
+
+    rdf = rdf.reindex(
```
lsimonsen, line 345: Aren't we re-creating the Data-Frame as a (business) daily frequency here (with stale values)? The observation frequency after down-sampling should be the frequency (such as one observation per monthly).

macrosynergy/management/utils/df_utils.py
```diff
@@ -300,6 +300,65 @@ def downsample_df_on_real_date(
     )
 
 
+def downsample_wide_df_on_real_date(
+    df: pd.DataFrame,
+    freq: str = "M",
+    agg: str = "mean",
+    ffill: bool = False,
```
lsimonsen, line 307: If the function ifs for down-sampling, we should never need to forward or backward fill.

macrosynergy/management/utils/df_utils.py
```diff
@@ -300,6 +300,65 @@ def downsample_df_on_real_date(
     )
 
 
+def downsample_wide_df_on_real_date(
```
lsimonsen, line 303: Is this function only to be used in historic portfolio volatility? Historically we have focused on convert frequency? Do we not have a frequency conversion function already (we calculate monthly returns)?

macrosynergy/management/utils/df_utils.py
```diff
@@ -300,6 +300,65 @@ def downsample_df_on_real_date(
...
+                df.index.max(),
+                freq="B",
+                name="real_date",
+            )
+        )
+    )
+    if ffill:
+        rdf = rdf.ffill(limit=FFILL_LIMITS[freq])
+    if bfill:
+        rdf = rdf.bfill(limit=FFILL_LIMITS[freq])
```
lsimonsen, line 358: Give above is return if `not (ffill or bfill)` and we specify only `assert not (ffill and bfill)`, should it not be something like:
```python
if ffill or bfill:
    rdf = rdf.reindex(
        pd.Series(
            pd.bdate_range(
                df.index.min(),
                df.index.max(),
                freq="B",
                name="real_date",
            )
        )
    )
if ffill:
    rdf = rdf.ffill(limit=FFILL_LIMITS[freq])
elif bfill:
    rdf = rdf.bfill(limit=FFILL_LIMITS[freq])
return rdf

File: macrosynergy/management/utils/df_utils.py [hunk 7]
```diff
@@ -901,3 +960,91 @@ def get_eops(
...
+        frequency.
+
+    :return <pd.DataFrame>: The estimated release frequency.
+    """
+    assert "timeseries" not in kwargs, "Argument `timeseries` is not allowed."
+    assert isinstance(df_wide, pd.DataFrame), "Argument `df_wide` must be a DataFrame."
+
+    return pd.Series(
+        [
+            estimate_release_frequency(
```
lsimonsen, line 1043: How come the hidden function is calling the non-hidden function? Isn't it usually the other way around?

macrosynergy/pnl/contract_signals.py
```diff
@@ -280,6 +351,7 @@ def contract_signals(
     hratios: Optional[str] = None,
     start: Optional[str] = None,
     end: Optional[str] = None,
+    est_freq: str = "M",
```
lsimonsen, line 354: Why does contract signals have an estimation frequency? Shouldn't it be a `rebalance_freq` if any?

lsimonsen (changes requested):
Still a few outstanding questions on this one.

macrosynergy/pnl/notional_positions.py
Magnus167: We use capitals because pandas uses it. `macrosynergy.management.utils.core._map_to_business_day_frequency` exists to provide that mapping consistently across package functions (those that have been refactored to do so anyway).

macrosynergy/management/utils/df_utils.py
Magnus167: the fills exist to pull data from whichever direction we will next use it if missing on an index.
e.g. If I downsample a daily dataframe to a monthy frequency, I end up with data only on the end of the month (30-Jan for example). 
Now when I want to use this data with weights, or in any method where I must consider the value for the next/previous data available data point, it leads to very large number of calls to pandas slicing/iloc/first_valid/last_valid_index

macrosynergy/management/utils/df_utils.py
Magnus167: We do, but it works on the quantamental format. Ultimately (many many refactors later), we want to many of the backend method to use wide DFs, with an overload for QDFs

macrosynergy/management/utils/df_utils.py
Magnus167: The hidden function in this case exists specifically to make the main function more readable and to provide clean compatibility for wide dataframes and single timeseries (`pd.Series`). 
This is very similar to cases where we use:
```python
def fmt(x: str):
    if isinstance(inputX, list):
        return list(map(fmt, inputX))
    return x.capitalize()
```
where the hidden function can be thought of as a fancy `map`

macrosynergy/pnl/contract_signals.py
Magnus167: Was uncertain which one this was. Will change it out soon

macrosynergy/management/utils/df_utils.py
lsimonsen: Got it - we should consider then mapping it out and maybe have [1] QDF and [2] WDF as two distinct types.

macrosynergy/management/utils/df_utils.py
lsimonsen: That seems to me a question of two part in the data-frame to have the frequency converted:

1. All non-last period (of down-sampling): so for monthly output targets, we can take everything before the last month and directly change frequency.
2. The last month in the sample, where we need to take the last observation, but we have two options: stock or flow variable. Stock variable is equivalent to the assumption of forward fill (last value), where as a flow variable (such as returns), needs to be the sum up to that date.

lsimonsen (changes requested):
See comments.

macrosynergy/management/utils/df_utils.py
Magnus167: Yes. For now, `downsample_wide_df_on_real_date` and `downsample_df_on_real_date` exist for that.
We will most likely be adding more functionality around wide dataframes, or atleast context based switching between the two formats

macrosynergy/management/utils/df_utils.py
Magnus167: How do we differentiate for (2) when contract signals are provided?

macrosynergy/management/utils/df_utils.py
Magnus167: fillna got deprecated in [pandas==2.1.0](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.fillna.html)

lsimonsen (approved):

PR #1528, opened 2024-03-05
`ProxyPNL` - Refactor constants.py and remove unused file

Magnus167:


lsimonsen:
@Magnus167 why move it out of `utils` and into management?

Magnus167:
It already exists as `msm.constants`, we want to keep one constants file across the package (hopefully)

lsimonsen (approved):

PR #1543, opened 2024-03-13
`ProxyPNL`, integrating `hist_portfolio_vol` into `notional_postions`

Magnus167:


macrosynergy/pnl/notional_positions.py
```diff
@@ -65,6 +65,94 @@ def _apply_slip(
...
+    sig_ident: str = f"_CSIG_{sname}"
+    _check_conts: Set = set([f"{contx}{sig_ident}" for contx in fids])
+    _found_conts: Set = set((df["cid"] + "_" + df["xcat"]).unique())
+    if not _check_conts.issubset(_found_conts):
+        raise ValueError(
+            f"Contract signals for all contracts not in dataframe. \n"
+            f"Missing: {_check_conts - _found_conts}"
+        )
+
+    return
```
lsimonsen, line 98: Do we need to explicitly specify `return` if it is empty?

macrosynergy/pnl/notional_positions.py
```diff
@@ -65,6 +65,94 @@ def _apply_slip(
...
+
+
+def _vol_target_positions(
+    df: pd.DataFrame,
+    sname: str,
+    fids: List[str],
+    dollar_per_signal: Numeric = 1.0,
+    vol_target: Numeric = 0.1,
```
lsimonsen, line 106: Minor point, but unless I am mistaken, we normally refer to volatility is percent, so `10` and not `0.1`.

lsimonsen (commented):
Looks good - now we just need the testing.

macrosynergy/pnl/notional_positions.py
Magnus167: Fixed

macrosynergy/pnl/notional_positions.py
```diff
@@ -79,14 +169,10 @@ def _leverage_positions(
     df_wide: pd.DataFrame = qdf_to_ticker_df(df=df)
     sig_ident: str = f"_CSIG_{sname}"
 
-    _check_conts: List[str] = [f"{contx}{sig_ident}" for contx in fids]
-    if not set(_check_conts).issubset(set(df_wide.columns)):
-        raise ValueError(
-            f"Contract signals for all contracts not in dataframe. \n"
-            f"Missing: {set(_check_conts) - set(df_wide.columns)}"
-        )
+    _found_conts: List[str] = [f"{contx}{sig_ident}" for contx in fids]
```
lsimonsen, line 172: Why call it found? Should it just be `_contracts` or `_select_contracts`? To me found implies you might not have found all of them.

lsimonsen (changes requested):
One small change request
* [ ] why the `_found_cont` name?

macrosynergy/pnl/notional_positions.py
Magnus167: Yes, this should be renamed. Earlier it was being used to actually check against a list of expected contracts, so I never renamed the variable

PR #1551, opened 2024-03-15
`ProxyPNL`, `historic_portfolio_vol()` - full/inf. lookback window

Magnus167:


macrosynergy/pnl/historic_portfolio_volatility.py
```diff
@@ -60,7 +60,7 @@ def _weighted_covariance(
 
     xnans, ynans = np.isnan(x), np.isnan(y)
     wmask = xnans | ynans
-    weightslen = min(sum(~wmask), lback_periods)
+    weightslen = min(sum(~wmask), lback_periods if lback_periods > 0 else len(x))
```
lsimonsen, line 63: Do we need to be explicit about only -1 is full look back? So basically:

```python
assert lback_periods == -1 or lback_periods > 0
```

Just to insure the user don't accidentally put in -21 days or similar?

macrosynergy/pnl/historic_portfolio_volatility.py
```diff
@@ -125,23 +125,32 @@ def _mask_nans(
     remove_zeros: bool,
 ) -> pd.DataFrame:
     """Mask NaNs in the dataframe"""
+    lback_periods = 0 if lback_periods < 0 else lback_periods
```
lsimonsen, line 128: Does this work with our `lback_periods = -1` for full sample or is the logic that 0 or less is full sample?

Magnus167: No, the only acceptable negative value is -1.

https://github.com/macrosynergy/macrosynergy/blob/d604e10c97a61d33744e86ba6611ef8e783a0482/macrosynergy/pnl/historic_portfolio_volatility.py#L205

macrosynergy/pnl/historic_portfolio_volatility.py
Magnus167: This is asserted earlier at https://github.com/macrosynergy/macrosynergy/blob/d604e10c97a61d33744e86ba6611ef8e783a0482/macrosynergy/pnl/historic_portfolio_volatility.py#L205

PR #1619, opened 2024-04-09
`ProxyPNL`, transaction costs calculation

Magnus167:


lsimonsen (approved):

macrosynergy/download/transaction_costs.py
```diff
@@ -66,5 +78,16 @@ def download_transaction_costs(
...
 
 
+def get_fids(df: QuantamentalDataFrame) -> list:
+    def repl(x: str, yL: List[str]) -> str:
+        for y in yL:
+            x = x.replace(y, "")
+        return x
+
+    return list(
+        set(df["cid"] + "_" + df["xcat"].apply(lambda x: repl(x, AVAILABLE_XCATS)))
```
lsimonsen, line 88: Do you not need to keep the `ctype`? Also for the AVAILABLE XCATS it is `<ctype><cost>_<stat>`

lsimonsen (changes requested):
Not sure about the `ExampleAdapter` and a minor comment in the download function. Otherwise looks good.

macrosynergy/download/transaction_costs.py
Magnus167: I think I have simply misnamed the variable;
It should be `AVAILABLE_CCATS` (contract categories) of the form:
`<ctype><cost>_<size>`
complementing `<cid>_<ctype><rtype>_<...`

macrosynergy/download/transaction_costs.py
```diff
@@ -32,6 +33,17 @@ def _request_wrapper(url: str, verbose: bool = True, **kwargs) -> str:
     raise requests.exceptions.RequestException(f"Failed to download {url}")
 
 
+AVAILABLE_CTYPES = ["FX", "IRS", "CDS"]
+AVAIALBLE_COSTS = ["BIDOFFER", "ROLLCOST", "SIZE"]
+AVAILABLE_STATS = ["MEDIAN", "90PCTL"]
+AVAILABLE_XCATS = [
+    f"{c}_{t}_{s}"
```
lsimonsen, line 40: xcat format for transaction costs: `f"{c}{t}_{s}"`

macrosynergy/download/transaction_costs.py
lsimonsen: Example filter `"AUD_FXBIDOFFER_MEDIAN".replace("FXBIDOFFER_MEDIAN", "") => "AUD"`

lsimonsen: Correct output should be `AUD_FX`.

macrosynergy/pnl/transaction_costs.py
```diff
@@ -0,0 +1,255 @@
...
+        for txn_ticker in [tcost_n, rcost_n, size_n, tcost_l, rcost_l, size_l]
+    ]
+    found_tickers = list(set(df["cid"] + "_" + df["xcat"]))
+    if not set(expected_tickers).issubset(set(found_tickers)):
+        raise ValueError(
+            "The dataframe is missing the following tickers: "
+            + ", ".join(set(expected_tickers) - set(found_tickers))
+        )
+
+    return None
```
lsimonsen, line 43: Isn't the `return None` statement redundant? Given it is the default behaviour if no return is specified.

macrosynergy/pnl/transaction_costs.py
```diff
@@ -0,0 +1,255 @@
...
+        size_n="SIZE_MEDIAN",
+        tcost_l="BIDOFFER_90PCTL",
+        rcost_l="ROLLCOST_90PCTL",
+        size_l="SIZE_90PCTL",
+    )
+
+    @classmethod
+    def download(self) -> "TransactionCosts":
+        df = download_transaction_costs()
+        return TransactionCosts(df=df, fids=get_fids(df), **self.DEFAULT_ARGS)
```
lsimonsen, line 124: Shouldn't this be:
```python
@classmethod
def download(cls) -> "TransactionCosts":
    df = download_transaction_costs()
    return cls(df=df, fids=get_fids(df), **self.DEFAULT_ARGS)
```
to follow standard notion?

lsimonsen: Also, personally I think it is more readable if defined after the `__init__` method, so you first give the user the standard way of instantiating the class, and then secondly the alternative (`download`).

lsimonsen: I like it though - even as I prefer downloading the data separately.

macrosynergy/pnl/transaction_costs.py
```diff
@@ -0,0 +1,255 @@
...
+        tcost_n: str,
+        rcost_n: str,
+        size_n: str,
+        tcost_l: str,
+        rcost_l: str,
+        size_l: str,
+    ) -> None:
+        self.sparse_costs = SparseCosts(df)
+        check_df_for_txn_stats(
+            df, fids, tcost_n, rcost_n, size_n, tcost_l, rcost_l, size_l
```
lsimonsen, line 139: I would strongly encourage we use the ability of python to have named arguments, so we know they map up correction to their positions.
```python
check_df_for_txn_stats(
            df, fids=fids, tcost_n=tcost_n, ...
)

File: macrosynergy/pnl/transaction_costs.py [hunk 6]
```diff
@@ -0,0 +1,255 @@
...
+        trade_size: Number,
+        median_size: Number,
+        median_cost: Number,
+        pct90_size: Number,
+        pct90_cost: Number,
+    ) -> Number:
+        # just as an example
+        u = trade_size * median_cost * median_cost / (median_size * median_size)
+        v = trade_size * pct90_cost * pct90_cost / (pct90_size * pct90_size)
+        return (u + v) / 2
```
lsimonsen, line 209: Equivalent to:
```python
u = median_cost * median_cost / (median_size * median_size)
v = pct90_cost * pct90_cost / (pct90_size * pct90_size)
avg_costs = (u+ v) /2
return trade_size * avg_costs
```
unless I am mistaken.

lsimonsen: Why cost and sized squared? Wouldn't this be better:

```python
u = median_cost / median_size
v = pct90_cost / pct90_size 
avg_costs = (u+ v) /2
return trade_size * avg_costs
```

lsimonsen (changes requested):
Looks good @Magnus167. A few minor comments and suggestions, otherwise I think we are good to go.

macrosynergy/pnl/transaction_costs.py
Magnus167: Mostly because it was just an example.
https://github.com/macrosynergy/macrosynergy/blob/adfea08c26305b4ecbe03ea326d51710c380ce9a/macrosynergy/pnl/transaction_costs.py#L206 

Have updated it with your example:
> 
> ```python
> u = median_cost / median_size
> v = pct90_cost / pct90_size 
> avg_costs = (u+ v) /2
> return trade_size * avg_costs
> ```

macrosynergy/pnl/proxy_pnl_calc.py
```diff
@@ -7,31 +7,174 @@
...
+        # for each column warns for dates of nas
+        for col in dfx.columns:
+            nas_idx = dfx[col].loc[dfx[col].isna()]
+            if not nas_idx.empty:
+                print(
+                    f"Warning: Series {col} has NAs at the following dates: {nas_idx.index}"
+                )
+
+    # Get the diff index for positions
+    pos_diff_index: pd.DatetimeIndex = get_diff_index(df_wide=pivot_pos)
```
lsimonsen, line 71: How is the black-listing handled in positions? Are we setting intra-periods N/A to zero if black-listed? We need to make sure only to rebalance for N/A when we go from N/A (or zero) to a value.

macrosynergy/pnl/proxy_pnl_calc.py
```diff
@@ -7,31 +7,174 @@
...
+    start: str = pivot_pos.first_valid_index()
+    end: str = pivot_pos.last_valid_index()
+
+    # List of rebal_dates - with end date as the last date of the PNL calc
+
+    rebal_dates = sorted(set(pos_diff_index) | {pd.Timestamp(end)})
```
lsimonsen, line 77: Why are we rebalancing on the end date? If it isn't on the first, we shouldn't rebalance.

Magnus167: The dnd date will end up as d2, so we're rebalancing up to but not including the end day.

lsimonsen (changes requested):
Review still in progress.

macrosynergy/pnl/proxy_pnl_calc.py
Magnus167: A better way to put it would be:
On the last rebalance day, as took a position, which continues to have a PNL until no more position data is available

macrosynergy/pnl/proxy_pnl_calc.py
Magnus167: In the actual entry point `proxy_pnl_calc`

lsimonsen: (black-listing was just an example of where I think we might have N/A's).

macrosynergy/pnl/proxy_pnl_calc.py
lsimonsen: Got it, that makes more sense. Thank you. Is it maybe worth to add the end-date only in the function when we rebalance (`rebal_dates + [pd.Timestamp(end)]`) to make it clear that `rebal_dates` are dates of rebalancing.

macrosynergy/pnl/proxy_pnl_calc.py
```diff
@@ -7,31 +7,174 @@
...
+
+    return_df_cols = pivot_pos.columns.tolist()
+    pnl_df = pd.DataFrame(index=pd.bdate_range(start, end), columns=return_df_cols)
+
+    return pnl_df, pivot_pos, pivot_returns, rebal_dates
+
+
+def pnl_excl_costs(
+    df: QuantamentalDataFrame,
+    spos: str,
```
lsimonsen, line 106: What does `spos` stand for? Signal positions? Should it be `npos` to make it comparable to notional positions?

macrosynergy/pnl/proxy_pnl_calc.py
Magnus167: does make sense, changing now

macrosynergy/pnl/proxy_pnl_calc.py
```diff
@@ -7,31 +7,174 @@
...
+) -> pd.DataFrame:
+
+    pnl_df, pivot_pos, pivot_returns, rebal_dates = _prep_dfs_for_pnl_calcs(
+        df=df, spos=spos, rstring=rstring
+    )
+
+    # between each rebalancing date
+    for dt1, dt2 in zip(rebal_dates[:-1], rebal_dates[1:]):
+        curr_pos: pd.Series = pivot_pos.loc[dt1]
+        curr_rets: pd.DataFrame = pivot_returns.loc[dt1:dt2]
```
lsimonsen, line 118: Question: will this include `dt2` in the returns? Goes back to the inclusion of `end` in the rebalance dates. Returns are end-of-day returns, so [1] need to make sure whether they are earned on the day position is taken, or at the end of the period. In either case, for the last period, we should make sure the returns are included for `dt2`.

macrosynergy/pnl/proxy_pnl_calc.py
```diff
@@ -7,31 +7,174 @@
...
+        cumprod_rets: pd.Series = (1 + curr_rets).cumprod()
+        pnl_df.loc[dt1:dt2] = curr_pos * cumprod_rets
+
+    # sum cols, ignore nans
+    pnl_df[spos + pnl_name] = pnl_df.sum(axis=1, skipna=True)
```
lsimonsen, line 123: Where do we get N/A's from?

macrosynergy/pnl/proxy_pnl_calc.py
```diff
@@ -7,31 +7,174 @@
...
+
+    for dt1, dt2 in zip(rebal_dates[:-1], rebal_dates[1:]):
+        curr_pos: pd.Series = pivot_pos.loc[dt1]
+        next_pos: pd.Series = pivot_pos.loc[dt2]
+        curr_rets: pd.DataFrame = pivot_returns.loc[dt1:dt2]
+        cumprod_rets: pd.Series = (1 + curr_rets).cumprod()
+        pnl_df.loc[dt1:dt2] = curr_pos * cumprod_rets
+        avg_pos_size = curr_pos.mean()
+        delta_pos = sum(abs(next_pos.values - curr_pos.values))
+        ...
```
lsimonsen, line 167: Does Python accept this as incomplete (but functional) code? I.e. can we merge it in with this notation?

macrosynergy/pnl/proxy_pnl_calc.py
```diff
@@ -107,3 +250,45 @@ def proxy_pnl_calc(
...
+        size_l=size_l,
+    )
+
+    # Create the wide dataframe
+    df_wide = qdf_to_ticker_df(df=df)
+
+
+if __name__ == "__main__":
+    ...
+    dfxcn = pd.read_pickle("data/dfxcn.pkl")
```
lsimonsen, line 300: Should we leave pickled data? If we are adding data to it, I would propose at least as a `CSV` so it is humanly readable. It is a very small dataset.

lsimonsen (changes requested):
A few minor outstanding questions.

macrosynergy/pnl/proxy_pnl_calc.py
Magnus167: yes, the code will run. in this particular case, the elllipsis is just interpreted as `pass`

macrosynergy/pnl/proxy_pnl_calc.py
Magnus167: generally from upstream functions. In more than a few cases, data is simply not available for some tickers

macrosynergy/pnl/proxy_pnl_calc.py
Magnus167: it's Strategy_Name + Positions_Name, i.e, in `contract_signals(sname='STRAT', ...)` and in `notional_positions(pname='POS')`. So SPOS for this is STRAT_POS.
I think it exists solely to allow running multiple variations of contract signals and notional positions and check their PNLs

macrosynergy/pnl/proxy_pnl_calc.py
Magnus167: I don't think so... I'm currently using it so I can "hand-off" from the notebooks

lsimonsen (approved):

PR #1663, opened 2024-04-24
`ProxyPNL`, integrating `TransactionCosts` object into `proxy_pnl_calc`

Magnus167:


macrosynergy/pnl/proxy_pnl_calc.py
```diff
@@ -330,40 +330,79 @@ def apply_trading_costs(
             f"_{spos}_{tc_name}", ""
         )
 
-        out_df[pnl_col] = out_df[pnl_col] - tc_wide_df[tc_col]
+        out_df[pnl_col] = out_df[pnl_col].sub(tc_wide_df[tc_col], fill_value=0)
```
lsimonsen, line 333: This looks dangerous to me. So if there is an N/A in the transaction cost data frame we just subtract zero? Needs to make sure it either throws and error or fills in N/A values if there is no transaction costs.

lsimonsen (changes requested):
Need to confirm behaviour of N/A values in transaction costs.

macrosynergy/pnl/proxy_pnl_calc.py
Magnus167: The txn costs dataframe comes in as a dense DF with transaction costs being applied solely on rebalance dates. this is also checked here:
https://github.com/macrosynergy/macrosynergy/blob/e7c90e7782b0ce38464abb276112a7094cd9e86a/macrosynergy/pnl/proxy_pnl_calc.py#L288


However, `out_df` has all the dates (superset of rebalance dates). When multiplied, the dates with no entry on transaction costs DF (any date except rebals.) get their values set to NaN. This is to solely to avoid that and treat all nans as 0 when subtracting.

Magnus167: example:
image

PR #1708, opened 2024-05-13
mainly edits docstrings

rsueppel:


rsueppel:
__init__

contract_cids => cid
contract_return => xcat
market_cid + market_return => benchmark_return



estimate_beta

neg_mean_abs_market_corr => neg_mean_abs_corr: negative mean absolute correlation (with the benchmark return)
=> where is the period set?

oos_period
=>
est_freq: period of sequential model selections and parameter estimations. Can be any of "d", "w", "m", or "q". Model and parameters are estimated at the end of the period and applied for hedged returns throughout the next period. Default is "m". 
+
:param <Bool> iis: initial in-sample estimates. If set to TRUE the returns for the initial minimum sample are hedged based on the in-sample beta estimate. This is not suitable for checks of the quality of the hedging. Default is FALSE.

PR #1791, opened 2024-06-18
`ProxyPNL`, bugfix for VCV/NA checks

Magnus167:


macrosynergy/pnl/historic_portfolio_volatility.py
```diff
@@ -170,6 +170,119 @@ def get_max_lookback(lb: int, nt: float) -> int:
...
+
+def _calc_vol_tuple(
+    vcv_df: pd.DataFrame,
+    signals: pd.DataFrame,
+    date: pd.Timestamp,
+) -> Tuple[pd.Timestamp, float]:
+    s = signals.loc[date, :].copy()
+
+    # reduce s to vcv columns
+    s = s[s.index.isin(vcv_df.columns)]
```
lsimonsen, line 225: This looks like something that should be an error rather than just filter it out? What is the logic from dropping the signals?

Magnus167: by this stage in the program flow, if the VCV matrix has NAs - we have verified either the signal itself is NA/0s, or that the VCV calculation is not possible for that window (due to data availability - NA/0s). Since a calculation cannot be done for a signal with missing VCV data, we can simply drop those columns

macrosynergy/management/utils/sparse.py
```diff
@@ -90,7 +90,8 @@ def create_delta_data(
 
     if not isinstance(df, QuantamentalDataFrame):
         raise ValueError("`df` must be a QuantamentalDataFrame")
-
+    if not isinstance(return_density_stats, bool):
```
lsimonsen: @Magnus167, did we accidentally merge in changes from the `sparse.py` script in another PR? I guess once we merge in the other changes to dev it should be fixed? Or is it a matter of merging in the develop branch?

macrosynergy/pnl/historic_portfolio_volatility.py
```diff
@@ -170,6 +170,125 @@ def get_max_lookback(lb: int, nt: float) -> int:
...
+    )
+
+    return vcv_df
+
+
+def _calc_vol_tuple(
+    vcv_df: pd.DataFrame,
+    signals: pd.DataFrame,
+    date: pd.Timestamp,
+    available_cids: List[str],
```
lsimonsen, line 221: Where do we define `available_cids`? Shouldn't this be given by the signals?

macrosynergy/pnl/historic_portfolio_volatility.py
```diff
@@ -210,69 +329,62 @@ def _calculate_portfolio_volatility(
...
-        dict_vcv: Dict[str, pd.DataFrame] = {
-            freq: estimate_variance_covariance(
-                piv_ret=_downsample_returns(window_df, freq=freq).iloc[
-                    -get_max_lookback(lb=lb, nt=nan_tolerance) :
-                ],
-                lback_periods=lb,
-                remove_zeros=remove_zeros,
-                weights_func=weights_func,
-                half_life=hl,
+        avails = first_starts[first_starts <= td].index.tolist()
```
lsimonsen, line 342: We need to check there is no signals associated with not available cross sections in prior sets of the sample.

macrosynergy/pnl/historic_portfolio_volatility.py
```diff
@@ -210,69 +329,62 @@ def _calculate_portfolio_volatility(
...
     pvol = pd.DataFrame(
         list_pvol,
         columns=["real_date", portfolio_return_name],
     ).set_index("real_date")
 
-    vcv_df = pd.concat(list_vcv, axis=0)  # add to cls.vcv
+    vcv_df_long = pd.concat(list_vcv, axis=0)  # add to cls.vcv
 
-    vcv_df["helper"] = vcv_df[["fid1", "fid2", "real_date"]].apply(
+    vcv_df_long["helper"] = vcv_df_long[["fid1", "fid2", "real_date"]].apply(
```
lsimonsen, line 377: This is on me, but lets find a better naming convention than `fid1` and `fid2`, maybe `fid_rows` and `fid_cols`?

lsimonsen (changes requested):
Minor changes and questions - but otherwise looks good to merge.

macrosynergy/management/utils/sparse.py
Magnus167: This is from merging dev back in I believe

macrosynergy/pnl/historic_portfolio_volatility.py
lsimonsen: Sounds reasonable, but requires the check to be in place that the signal is NA/0 before we drop it.

lsimonsen (approved):

PR #1796, opened 2024-07-02
Feature: Integration test about requirements and change minimum version of requirements

sandresen1:
Change this so instead of having a requirements-min.txt file specified it is dynamically created in the workflow, meaning it is always up to date with pyproject.toml and requirements.txt


emhbrine (dismissed):

emhbrine (approved):

pyproject.toml
Magnus167, line 1: Can we also merge the optional dependencies into 2 broad groups?
`[dev]` or `[test]` for all docs+testing+etc. and another for `[all]`. The `[all]` list should contain all test/dev requirements and everything required to start a notebook in jupyter (ipykernel, jupyter, notebook, etc.)

PR #1809, opened 2024-07-12
Feature: Information State Changes

lsimonsen:
Key functionality added:

1. Extract information state change per ticker, which is a sparser object and takes into account the non-linear time arrival of information flows, in a new `InformationStateChanges` class.
2. Allows us to normalise these changes in information states relative to the first releases of an underlying observation period, which is normally a more volatile change, than a revision to an existing print.
3. From the class we can use the method of `to_qdf` to recover the canonical daily `QuantamentalDataFrame` format of the data.
4. We have also added temporal aggregator functions, for calculating a variety of moving averages and sums (including for specific periods).


macrosynergy/management/utils/sparse.py
```diff
@@ -0,0 +1,821 @@
...
+    @staticmethod
+    def abs(s: pd.Series, min_periods: int, **kwargs) -> pd.Series:
+        """
+        Calculate the absolute standard deviation of a Series.
+
+        :param <pd.Series> s: The Series to calculate the absolute standard deviation for.
+        :param <int> min_periods: The minimum number of periods required for the calculation.
+        :return: The absolute standard deviation of the Series.
+        """
+        return s.abs().expanding(min_periods=min_periods).mean()
```
emhbrine, line 176: Does this calculate the absolute standard deviation of a Series?

macrosynergy/management/utils/sparse.py
```diff
@@ -0,0 +1,821 @@
...
+        """
+        Calculate the exponentially weighted absolute standard deviation of a Series.
+
+        :param <pd.Series> s: The Series to calculate the exponentially weighted absolute
+            standard deviation for.
+        :param <int> halflife: The halflife of the exponential weighting.
+        :param <int> min_periods: The minimum number of periods required for the calculation.
+        :return: The exponentially weighted absolute standard deviation of the Series.
+        """
+        return s.abs().ewm(halflife=halflife, min_periods=min_periods).mean()
```
emhbrine, line 203: Please correct me if I'm wrong, but this also looks like it's calculating the mean instead of std.

macrosynergy/management/utils/sparse.py
emhbrine: @lsimonsen cleared up my confusion, but I suggest we go with `average absolute deviation` in the docstring

emhbrine (approved):

PR #1825, opened 2024-07-16
Feature: Adding `get_releases` and `to_json` methods for `InformationStateChanges` class

lsimonsen:
Add three methods to the class:

* `to_dict`: creates a dictionary of the sparse data for a specific ticker.
* `to_json`: creates a JSON string from the output of `to_dict`.
* `get_releases`: extract the data release between two dates (defaults is all release from last business day till today).

### DIFF
--- a/macrosynergy/management/utils/sparse.py
+++ b/macrosynergy/management/utils/sparse.py
@@ -10,6 +10,7 @@ from macrosynergy.management.utils import (
     concat_single_metric_qdfs,
     get_cid,
     get_xcat,
+    is_valid_iso_date,
 )
 from macrosynergy.management.types import QuantamentalDataFrame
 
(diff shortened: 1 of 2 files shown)


macrosynergy/management/utils/sparse.py
```diff
@@ -762,6 +763,62 @@ def to_qdf(
...
+    def get_releases(
+        self,
+        from_date: pd.Timestamp = pd.Timestamp.today().normalize() - pd.offsets.BDay(1),
+        to_date: pd.Timestamp = pd.Timestamp.today().normalize(),
+        excl_xcats: List[str] = None,
+    ) -> pd.DataFrame:
+        store = []
+        for k, v in self.items():
+            real_date = v.index.max()
+            s = v[v.index == real_date]
```
Magnus167, line 796: Should this be changed to iterate over each unique release in [from_date, to_date] instead of just the last one?

Magnus167 (approved):

PR #2082, opened 2024-11-06
Update documentation to replace 'xcats' with 'categories', and 'cids' with 'cross-section' where relevant

Magnus167:


sandresen1 (dismissed):

macrosynergy/panel/granger_causality_test.py
```diff
@@ -40,13 +40,13 @@ def granger_causality_test(
...
-    Granger Causality Test is a pairwise test, only two tickers can be specified. Specify
-    either `tickers` or `cids` & `xcats`. When specifying `cids` & `xcats`, the user may
-    input one `cid` and two `xcats`; or two `cids` and one `xcat` to yield two tickers.
-    The function forms the list of tickers from the `cids` and `xcats` such that the
-    order of the formed tickers is preserved. The order of the tickers is important as
-    the first ticker is the one that is tested to Granger cause the second ticker. The
-    function tests whether the first ticker Granger causes the second ticker.
+    Granger Causality Test is a pairwise test, only two tickers can be specified.
+    The user may specify two tickers directly using the `tickers` parameter, or specify
+    one or two cross-sections and one or two extended categories using the `cids` and
```
emhbrine, line 45: I think the meaning has changed here.

From 

one `cid` and two `xcats`; or two `cids` and one `xcat`

to 

one or two `cid` and one or two `xcats`

It sounds like you could pass two cids and two xcats

emhbrine: Is that how it should be?

Magnus167: I agree, will change this to read better

emhbrine (approved):

PR #2261, opened 2025-03-10
Feature: Implement `adjust_weights` functionality

Magnus167:
Introduce a new feature that modifies standard weights based on signal zn-scores.

### DIFF
--- a/macrosynergy/panel/adjust_weights.py
+++ b/macrosynergy/panel/adjust_weights.py
@@ -201,8 +201,8 @@ def adjust_weights(
     if not isinstance(df, QuantamentalDataFrame):
         raise TypeError("df must be a QuantamentalDataFrame")
 
-    df = QuantamentalDataFrame(df)
-    result_as_categorical = df.InitializedAsCategorical
+    df: QuantamentalDataFrame = QuantamentalDataFrame(df)
+    result_as_categorical: bool = df.InitializedAsCategorical
 
     if cids is None:
         cids = df["cid"].unique().tolist()
--- a/macrosynergy/panel/adjust_weights.py
+++ b/macrosynergy/panel/adjust_weights.py
@@ -130,8 +130,18 @@ def split_weights_adj_zns(
     df_weights_wide = QuantamentalDataFrame(df_weights).to_wide()
     df_adj_zns_wide = QuantamentalDataFrame(df_adj_zns).to_wide()
 
-    # convert the weights to absolute values (no negative weights in this use case)
-    df_weights_wide = df_weights_wide.abs()
+    # cannot tolerate negative weights
+    if any(df_weights_wide[~df_weights_wide.isna()].lt(0).any()):
+        na_frame = QuantamentalDataFrame.from_wide(
+            df_weights_wide[
+                df_weights_wide[~df_weights_wide.isna()].lt(0).any(axis="columns")
+            ]
+        )
+
+        na_frame = na_frame[na_frame["value"] < 0]
+        raise ValueError(
+            f"Negative weights found in the dataframe. Please check the following data:\n{na_frame}"
+        )
 
     combined_index = df_weights_wide.index.union(df_adj_zns_wide.index)
     df_weights_wide = df_weights_wide.reindex(combined_index)
@@ -292,9 +302,10 @@ if __name__ == "__main__":
     df = make_test_df(xcats=["weights", "adj_zns"], cids=["cid1", "cid2", "cid3"])
     dfb = make_test_df(xcats=["some_xcat", "other_xcat"], cids=["cid1", "cid2", "cid4"])
 
-    nan_mask = np.random.rand(len(df)) < 0.25
-    df.loc[nan_mask, "value"] = np.nan
-
+    # nan_mask = np.random.rand(len(df)) < 0.01
+    # df.loc[nan_mask, "value"] = np.nan
+    # nan_mask = np.random.rand(len(df)) < 0.1
+    # df.loc[nan_mask, "value"] *= -1
     df = pd.concat([df, dfb], axis=0)
 
     def sigmoid(x, amplitude=1.0, steepness=1.0, midpoint=0.0):
--- a/macrosynergy/panel/adjust_weights.py
+++ b/macrosynergy/panel/adjust_weights.py
@@ -160,7 +160,11 @@ def normalize_weights(df_weights_wide: pd.DataFrame) -> pd.DataFrame:
     pd.DataFrame
         DataFrame with normalized weights (sum of each row is 1).
     """
-    df_weights_wide = df_weights_wide.div(df_weights_wide.sum(axis=1), axis=0)
+    nan_mask = df_weights_wide.isna()
+    df_weights_wide = df_weights_wide[~nan_mask].div(
+        df_weights_wide[~nan_mask].abs().sum(axis=1),
+        axis=0,
+    )
     return df_weights_wide
 
 
--- a/macrosynergy/panel/adjust_weights.py
+++ b/macrosynergy/panel/adjust_weights.py
@@ -97,7 +97,7 @@ def adjust_weights_backend(
         cid_weight = df_weights_wide[cid]
         cid_adj_zns = df_adj_zns_wide[cid]
 
-        result = cid_adj_zns * cid_weight.apply(method) * param
+        result = cid_weight * cid_adj_zns.apply(method) * param
         dfw_result[cid] = result
 
     return dfw_result
--- a/macrosynergy/panel/adjust_weights.py
+++ b/macrosynergy/panel/adjust_weights.py
@@ -236,25 +236,25 @@ def adjust_weights(
 
     df_weights_wide = normalize_weights(df_weights_wide)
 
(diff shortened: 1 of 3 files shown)


macrosynergy/panel/adjust_weights.py
```diff
@@ -0,0 +1,248 @@
...
+        DataFrame with weights in wide format.
+
+    df_adj_zns_wide : pd.DataFrame
+        DataFrame with adjustment factors in wide format.
+
+    method : Callable
+        Function that will be applied to the weights to adjust them. This function must
+
+    param : Number
+        Parameter that will be passed to the method function.
```
lsimonsen, line 83: Currently `param` isn't being passed to the `method` function, but scaling it:

Current (multiplication):
```python
result = cid_weight * cid_adj_zns.apply(method) * param
```
Passing it:
```python
result = cid_weight * cid_adj_zns.apply(method, *param)
```

lsimonsen: If it is a scale, shouldn't it be called a scale?

macrosynergy/panel/adjust_weights.py
```diff
@@ -0,0 +1,248 @@
...
+    assert set(df_weights_wide.columns) == set(df_adj_zns_wide.columns)
+    cids = sorted(df_weights_wide.columns)
+    dfw_result = pd.DataFrame(index=df_weights_wide.index)
+
+    for cid in cids:
+        cid_weight = df_weights_wide[cid]
+        cid_adj_zns = df_adj_zns_wide[cid]
+
+        result = cid_weight * cid_adj_zns.apply(method) * param
+        dfw_result[cid] = result
```
lsimonsen, line 100: Would it not be more efficient to apply it verctorised?
```python
df_weights_wide * method( df_adj_zns_wide) * param
```
instead of the for-loop.

macrosynergy/panel/adjust_weights.py
```diff
@@ -0,0 +1,258 @@
...
+        List of cids to adjust. If None, all cids will be adjusted. Default is None.
+    adj_name : str, optional
+        Name of the resulting xcat. Default is "ADJWGT".
+    """
+
+    if not isinstance(df, QuantamentalDataFrame):
+        raise TypeError("df must be a QuantamentalDataFrame")
+
+    df = QuantamentalDataFrame(df)
+    result_as_categorical = df.InitializedAsCategorical
```
lsimonsen, line 205: Add type-hints to make it easier for the reader to navigate:
```python
df: QuantamentalDataFrame = QuantamentalDataFrame(df)
result_as_categorical: bool = df.InitializedAsCategorical
``

File: macrosynergy/panel/adjust_weights.py [hunk 4]
```diff
@@ -0,0 +1,258 @@
...
+
+    if cids is None:
+        cids = df["cid"].unique().tolist()
+
+    check_types(weights, adj_zns, method, params, cids)
+
+    df, r_xcats, r_cids = reduce_df(
+        df, cids=cids, xcats=[weights, adj_zns], intersect=True, out_all=True
+    )
```
lsimonsen, line 214: What happens if the signal are not updated but the weights are? Should we throw an error?

Obviously one of them might have a longer history, but at the point today, we need to make sure they match? Also, do we have checks that there are no gaps in the middle (between first and last)?

macrosynergy/panel/adjust_weights.py
```diff
@@ -0,0 +1,258 @@
...
+
+    check_missing_cids_xcats(weights, adj_zns, cids, r_xcats, r_cids)
+
+    df_weights_wide, df_adj_zns_wide = split_weights_adj_zns(df, weights, adj_zns)
+
+    df_weights_wide = normalize_weights(df_weights_wide)
```
lsimonsen, line 220: We are we normalizing ahead of the adjustment? I can understand the normalisation below (line 232), but isn't this one redundant? It is just a scaler applied.

macrosynergy/panel/adjust_weights.py
```diff
@@ -0,0 +1,258 @@
...
+
+    dfw_result = normalize_weights(dfw_result)
+
+    if dfw_result.isna().all().all():
+        raise ValueError(
+            "The resulting DataFrame is empty. Please check the input data,"
+            " the method function, and it's parameters."
+        )
+
+    dfw_result.columns = list(map(lambda x: f"{x}_{adj_name}", dfw_result.columns))
```
lsimonsen, line 240: To simplify
```python
dfw.result.columns += adj_name
```
instead of
```python
dfw_result.columns = list(map(lambda x: f"{x}_{adj_name}", dfw_result.columns))
```

lsimonsen (changes requested):
A few smaller points and suggestions, but otherwise looks reasonable to me.

@GlennRe9  anything from your side?

macrosynergy/panel/adjust_weights.py
```diff
@@ -0,0 +1,258 @@
...
+    Returns
+    -------
+    pd.DataFrame
+        DataFrame with the adjusted weights.
+    """
+
+    assert set(df_weights_wide.columns) == set(df_adj_zns_wide.columns)
+    assert set(df_weights_wide.index) == set(df_adj_zns_wide.index)
+
+    dfw_result = df_weights_wide * df_adj_zns_wide.apply(method, **params)
```
lsimonsen, line 87: My one worry is the use of `.apply` which to my knowledge is notoriously slow.

Could we apply something like [functools partial](https://www.learnpython.org/en/Partial_functions) to it and then use it directly?

```python
from functools import partial
fn = partial(method, **params)
dfw_result = df_weights_wide * fn(df_adj_zns_wide)
```
or for the last line `dfw_result = df_weights_wide * df_adj_zns_wide.map(fn)`.

macrosynergy/panel/adjust_weights.py
Magnus167: removed

macrosynergy/panel/adjust_weights.py
```diff
@@ -0,0 +1,270 @@
...
+    df_weights_wide : pd.DataFrame
+        DataFrame with weights in wide format. (one column per cid)
+
+    Returns
+    -------
+    pd.DataFrame
+        DataFrame with normalized weights (sum of each row is 1).
+    """
+    df_weights_wide = df_weights_wide.div(
+        df_weights_wide.sum(axis="columns"), axis="index"
```
sandresen1, line 159: abs this

macrosynergy/panel/adjust_weights.py
lsimonsen: How is the performance of `.map(method, **params)`?

```python
dfw_result = df_weights_wide * df_adj_zns_wide.map(method, **params)
```

macrosynergy/panel/adjust_weights.py
```diff
@@ -0,0 +1,306 @@
...
+    """
+
+    df_weights = df.loc[df["xcat"] == weights]
+    df_adj_zns = df.loc[df["xcat"] == adj_zns]
+
+    df_weights_wide = QuantamentalDataFrame(df_weights).to_wide()
+    df_adj_zns_wide = QuantamentalDataFrame(df_adj_zns).to_wide()
+
+    # convert the weights to absolute values (no negative weights in this use case)
+    df_weights_wide = df_weights_wide.abs()
```
lsimonsen, line 130: @Magnus167 to me this should be a ValueError, rather than force it to be positive. If we have a value of -0.5, we don't know if the weight is actually 0.5 or if it is a rounding error.

@sandresen1 was this your point about absolute as well?

In general, we should have a check for:

1. df_weights all N/A or positive (between either 0 to 100 or 0 to 1).
2. The transformed weights being similar.

macrosynergy/panel/adjust_weights.py
```diff
@@ -0,0 +1,306 @@
...
+    weights_missing_in_zns = set(df_weights_wide.columns) - set(df_adj_zns_wide.columns)
+    zns_missing_in_weights = [f"{c}_{adj_zns}" for c in zns_missing_in_weights]
+    weights_missing_in_zns = [f"{c}_{weights}" for c in weights_missing_in_zns]
+    all_missing = zns_missing_in_weights + weights_missing_in_zns
+    if all_missing:
+        raise ValueError(f"Missing tickers: {all_missing}")
+
+    # get the corresponding rows in zns
+    nan_zns_rows = df_adj_zns_wide.isna().all(axis="columns")
+    all_zero_zns_rows = (df_adj_zns_wide.fillna(0) == 0).all(axis="columns")
```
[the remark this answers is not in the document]
lsimonsen: Obviously in the opposite scenario of historically missing signals, we want to fill in the zero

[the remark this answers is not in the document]
Magnus167: The two checks are as follows:

1. `nan_zns_rows ← index of rows where all elements are NaNs`
2. `all_zero_zns_rows ← ephemerally fill all nans with 0, and get index of rows where all are 0s`

sandresen1 (approved):

lsimonsen (approved):

PR #2262, opened 2025-03-11
Feature: Add a seed argument to `make_qdf`

Magnus167:
Introduce a random seed for consistent output in the `make_qdf` function


rushilg99 (approved):

rushilg99 (approved):

rushilg99 (changes requested):
Ah I think it makes sense for the default to be None for the seed.

rushilg99 (approved):

PR #2582, opened 2026-03-13
Feature: blacklisting for multiple targets

scubastef:
* amend signal optimizer to support different blacklists for different target variables
* change how long format dataframes are created in BasePanelLearner when n_targets > 1
* add tests for long format dataframe calculations

### DIFF
--- a/macrosynergy/learning/sequential/base_panel_learner.py
+++ b/macrosynergy/learning/sequential/base_panel_learner.py
@@ -65,10 +65,12 @@ class BasePanelLearner(ABC):
         end : str, 
             End date for considered data in subsequent analysis in ISO 8601 format.
             Default is None i.e. the latest date in the dataframe.
-        blacklist : list, 
+        blacklist : dict or list[dict],
             Blacklisting dictionary specifying date ranges for which cross-sectional
             information should be excluded. The keys are cross-sections and the values
-            are tuples of start and end dates in ISO 8601 format. Default is None.
+            are tuples of start and end dates in ISO 8601 format. A list of dictionaries
+            can be passed as well, where the ith blacklisting dictionary corresponds to
+            the ith target. Default is None.
         freq : str, 
             Frequency of the data. Default is "M" for monthly.
         lag : int, 
@@ -145,44 +147,19 @@ class BasePanelLearner(ABC):
         self.drop_nas = drop_nas
 
         # Create long-format dataframe
-        if self.n_targets == 1:
-            df_long = (
-                categories_df(
-                    df=self.df,
-                    xcats=self.xcats,
-                    cids=self.cids,
-                    start=self.start,
-                    end=self.end,
-                    blacklist=self.blacklist,
-                    freq=self.freq,
-                    lag=self.lag,
-                    xcat_aggs=self.xcat_aggs,
-                )
-            )
-        else:
-            features_xcats = self.xcats[:-self.n_targets]
-            targets_xcats = self.xcats[-self.n_targets:]
-            dfs = []
-            for target in targets_xcats:
-                df_long = (
-                    categories_df(
-                        df=self.df,
-                        xcats=features_xcats + [target],
-                        cids=self.cids,
-                        start=self.start,
-                        end=self.end,
-                        blacklist=self.blacklist,
-                        freq=self.freq,
-                        lag=self.lag,
-                        xcat_aggs=self.xcat_aggs,
-                    )
-                )
-                dfs.append(df_long)
-            df_long = pd.concat(dfs, axis=1).sort_index()
-            # Filter out duplicate categories
-            df_features = df_long.iloc[:,:len(features_xcats)]
-            df_targets = df_long[targets_xcats]
-            df_long = pd.concat([df_features, df_targets], axis=1)
+        df_long = _create_long_format_df(
+            df=self.df,
+            xcats=self.xcats,
+            cids=self.cids,
+            n_targets=self.n_targets,
+            start=self.start,
+            end=self.end,
+            blacklist=self.blacklist,
+            freq=self.freq,
+            lag=self.lag,
+            xcat_aggs=self.xcat_aggs,
+        )
+
         if self.drop_nas:
             df_long = df_long.dropna().sort_index()
         else:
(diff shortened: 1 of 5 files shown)


macrosynergy/learning/sequential/base_panel_learner.py
```diff
@@ -1229,31 +1206,44 @@ def _check_init(
...
+            if not isinstance(blacklist, (dict, list)):
+                raise TypeError("The blacklist argument must be a dictionary or list.")
+
+            if isinstance(blacklist, list) and n_targets > 1 and len(blacklist) != n_targets:
+                raise ValueError(
+                    f"When blacklist is a list and n_targets > 1, blacklist must have "
+                    f"exactly n_targets ({n_targets}) elements."
+                )
+
+            blacklists = [blacklist] if isinstance(blacklist, dict) else blacklist
```
rushilg99: what about the case where n_targets = 1 and only a dict is passed

macrosynergy/learning/sequential/base_panel_learner.py
```diff
@@ -1229,31 +1206,44 @@ def _check_init(
...
+
+            for i, blacklist in enumerate(blacklists):
```
rushilg99: I think this check needs to be done separately. From a client perspective, if they enter a dictionary, its converted to a list and we return that the first blacklist needs to be a dict... but they only entered one, and it can create confusion

macrosynergy/learning/sequential/base_panel_learner.py
```diff
@@ -1792,3 +1782,83 @@ def _remove_results(self, conditions):
...
+) -> list[dict | None]:
+    if isinstance(blacklist, list):
+        return blacklist
+    return [blacklist] * n_targets
+
+
+def _create_long_format_df(
+    df: pd.DataFrame,
+    xcats: list[str],
+    cids: list[str] | None,
```
rushilg99: fine for now but we have a standard of using the typing module and writing type hints as Optional[List[str]] as an example. 

Just one for both of us to be aware of whenever a 2.0 package release is done.

macrosynergy/learning/sequential/base_panel_learner.py
```diff
@@ -1792,3 +1782,83 @@ def _remove_results(self, conditions):
...
+        cids=cids,
+        start=start,
+        end=end,
+        freq=freq,
+        lag=lag,
+        xcat_aggs=xcat_aggs,
+    )
+
+    if n_targets == 1:
+        return categories_df(df=df, xcats=xcats, blacklist=blacklist, **shared_kwargs)
```
rushilg99, line 1845: we converted blacklist to a list earlier. does this still work?

macrosynergy/learning/sequential/base_panel_learner.py
```diff
@@ -1792,3 +1782,83 @@ def _remove_results(self, conditions):
...
+def _resolve_blacklists(
+    blacklist: dict | list[dict] | None,
+    n_targets: int,
+) -> list[dict | None]:
+    if isinstance(blacklist, list):
+        return blacklist
+    return [blacklist] * n_targets
+
+
+def _create_long_format_df(
```
rushilg99, line 1819: please add a test to generate a long format dataframe, in all cases, and verify the output is as you expect

macrosynergy/learning/sequential/base_panel_learner.py
```diff
@@ -1792,3 +1782,83 @@ def _remove_results(self, conditions):
...
+
+    if blacklist is None:
+        return df
+
+    # 1. rows where all targets are part of a blacklist should be removed
+    # 2. rows where some targets are blacklisted and some not should stay
+    # as is except blacklisted target should be set to nan
+    # todo: point 2. begs the question of how would we work with missing
+    # y data when multiple blacklists exist? can't be sure if nan is from blacklist
+    # or is just missing
```
rushilg99: we don't need to distinguish between the two cases.

macrosynergy/learning/sequential/base_panel_learner.py
```diff
@@ -1792,3 +1782,83 @@ def _remove_results(self, conditions):
...
+
+    features_xcats = xcats[:-n_targets]
+    targets_xcats = xcats[-n_targets:]
+
+    dfs = (
+        categories_df(
+            df=df,
+            xcats=features_xcats + [target],
+            **shared_kwargs,
+        )
```
rushilg99, line 1855: why can't we pass the blacklists in here, the resulting dataframes are concatenated later.

scubastef: Imagine we had two targets and two blacklists. If we pass the blacklists directly to `categories_df` we will end up concatenating two data frames of different sizes (the first df1 will be missing rows associated with blacklist1 and df2 will be missing rows associated with blacklist2). 

The concatenation will still work, but it will fill in those missing dates with nans. So, we now have X data with nans when there actually exists data

macrosynergy/learning/sequential/base_panel_learner.py
scubastef: I've updated it anyway because I don't think `|` is valid in python 3.7

macrosynergy/learning/sequential/base_panel_learner.py
scubastef: Good catch. I've updated it to work and added `test_single_target_with_blacklist_as_list` as a test

macrosynergy/learning/sequential/base_panel_learner.py
scubastef: Have removed this

macrosynergy/learning/sequential/base_panel_learner.py
scubastef: `test_base_panel_learner.py` has quite a few tests for this function which should cover all cases

macrosynergy/learning/sequential/base_panel_learner.py
scubastef: Good point. I've made the checks separate depending on if a list or dictionary is passed by the user

macrosynergy/learning/sequential/base_panel_learner.py
scubastef: The dict would be turned into a list with just that dict and pass the checks. However, given your other comment about splitting out the checks, I've removed this now

PR #2624, opened 2026-04-14
Feature: Adjust model fit check

scubastef:
* use a sample of the input data to check model can be fitted

### DIFF
--- a/macrosynergy/learning/sequential/base_panel_learner.py
+++ b/macrosynergy/learning/sequential/base_panel_learner.py
@@ -1379,8 +1379,29 @@ class BasePanelLearner(ABC):
                 )
 
             # Check model runs successfully on X and y
+            if isinstance(models[key], Pipeline):
+                n_samples = max(getattr(step, "min_periods", 0) for step in models[key])
+            else:
+                n_samples = getattr(models[key], "min_periods", 0)
+            n_samples = max(2 * n_samples, 10)
+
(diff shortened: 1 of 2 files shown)


rushilg99:
we just need to test on the first training set (the smallest one) in the pipeline

it implicitly tests whether the configurations of signal optimizer are plausible in addition to the correctness of the model code

rushilg99:
TODO: move check to after we precompute the split indices

macrosynergy/learning/sequential/base_panel_learner.py
```diff
@@ -338,6 +338,17 @@ def run(
...
 
+        # Check models can be fitted on first train set
+        X_first = self.X.iloc[train_test_splits[0][0]]
+        y_first = self.y.iloc[train_test_splits[0][0]]
+        for key, model in models.items():
+            try:
+                models[key].fit(X_first, y_first)
+            except Exception as e:
+                raise RuntimeError(
+                    f"Model {key} cannot be fit on the given X and y: {e}"
```
rushilg99: Message can be more transparent. Not clear to the user what X and y here actually mean. We need a message to say that.

Since this is an important user facing tool, that's why we need to be conscientious with the error message.

rushilg99: Also, make sure this is covered by the test please

rushilg99 (approved):

PR #2742, opened 2026-08-18
Bugfix: Forward fill pnl signals by only one period 

rushilg99:


mikiinterfiore (changes requested):
It does not seem it f-fills only by 1. I do not see the 1 anywhere

The if / else can also be replaced by pd.Grouper("real_date", freq=<mapped frequencies>)

The weekly case is also wrong as a week can span two years so we should ensure continuity.

rushilg99:
Can be closed since we opted to amend the signal to add zeros when an asset shouldn't be traded rather than amend the PnL.

